PR #8, opened 2023-01-25
Major Functionality Additions

BSd3v:
- added `rowTransaction` which utilizes the function `applyTransaction()` from AG Grid, allowing for more flexible updates from within AG Grids own api 
- removed the need for function parsing to need an additional prop
- updated prop names to make more sense when going to execute a function
- allowed for functions to be passed in multiple columns
- can add custom components into the namespace `dashAgGridComponentFunctions` to return react components
- can add custom variables for use in functions and dropdowns via namespace `dashSharedVariables`


src/lib/components/AgGrid.react.js
```diff
@@ -161,6 +161,11 @@ DashAgGrid.propTypes = {
                 PropTypes.bool, PropTypes.arrayOf(PropTypes.object)
         ]),
 
+    /**
+     * If true, the internal method updateRows() will be called
+     */
+    enableUpdateRows: PropTypes.arrayOf(PropTypes.object),
```
alexcjohnson, line 167: I'm not sure I understand, can you give an example of how to use this? Eventually we want all these examples to be in the form of dash testing test cases, ie you set up an app and then interact with it and the test says what should change ("click on this button and the third row changes to have this new data in it but the other rows are unchanged") but until we get that pattern going maybe you can just add an app file into the `tests/` dir? (just don't start its name with `test`, so we know it'll need converting to a test case later)

BSd3v: @alexcjohnson, do you want me to keep the rest of the test as is, or just remove the other stuff and leave only the table?

alexcjohnson: you mean the other files in that dir? I'd leave those there for now, just add a new file that uses this feature and has a comment describing the actions you can do in the app and what their result is supposed to be.

BSd3v: Added a test version for adding, updating, and removing rows.

Adding rows gets glitchy due to possibly duplicated rows, which goes berserk.

BSd3v: @alexcjohnson, I've made quite a few adjustments, as well as testing using clientside_callbacks to see if it bucks.

src/lib/fragments/AgGrid.react.js
```diff
@@ -17,6 +17,7 @@ import 'ag-grid-community/styles/ag-grid.css';
 import 'ag-grid-community/styles/ag-theme-alpine.css';
 import 'ag-grid-community/styles/ag-theme-balham.css';
 import 'ag-grid-community/styles/ag-theme-material.css';
+import * as d3 from 'd3'
```
alexcjohnson, line 20: The whole of D3 is pretty big (~300kb minified), and most of it is not useful for us. The pieces I can see wanting are `d3-format`, `d3-time-format`, and `d3-time`. `d3-array`? Do these functions have access to enough data to make `d3-array` useful?

Everything else seems too specialized or related to DOM manipulation so not useful to include here.

src/lib/fragments/AgGrid.react.js
```diff
@@ -349,6 +386,10 @@ export default class DashAgGrid extends Component {
             gridColumnApi: params.columnApi,
         });
 
+        if (this.props.rowData) {
+            this.state.rowData = JSON.parse(JSON.stringify(this.props.rowData))
```
alexcjohnson, line 390: I see a lot of these cycle-through-JSON operations, I'd like to see if we can avoid them. The ones involving row data will be very expensive for large tables. I guess the issue is ag grid mutating the `rowData` object on user edits... check out https://github.com/ag-grid/ag-grid/issues/3230 and the StackOverflow Q&A in there for a different approach that I think will avoid the need for this deepcopy operation.

src/lib/fragments/AgGrid.react.js
BSd3v: I'll see if I can combine them, I think with the addition of the namespace functions for formatting, we should be fine to pair it down.

src/lib/fragments/AgGrid.react.js
BSd3v: I'll see what we can do. The issue I think with this is what happens when a custom valueSetter gets added. Also, I think we still have the issue of multiple manipulations.

If we can figure out a better way to copy the old data, we wont need this. And maybe that's what we should do, using componentDidUpdate, when I tried initially it was updating too fast I think.

package.json
```diff
@@ -28,6 +28,10 @@
         "ag-grid-community": ">=29.0.0",
         "ag-grid-enterprise": ">=29.0.0",
         "ag-grid-react": ">=29.0.0",
+        "d3-format": ">=3.1.0",
+        "d3-array": ">=3.2.2",
+        "d3-time": ">=3.1.0",
+        "d3-time-format": ">=4.1.0",
```
alexcjohnson, line 34:
```suggestion
        "d3-format": "^3.1.0",
        "d3-array": "^3.2.2",
        "d3-time": "^3.1.0",
        "d3-time-format": "^4.1.0",
```
The difference is `^` won't go up a major version so won't accidentally bring in breaking changes, whereas `>=` can bump major versions.

src/lib/components/AgGrid.react.js
```diff
@@ -101,9 +101,9 @@ DashAgGrid.propTypes = {
 
 
     /**
-    * Enables the use of html for the whole table, this is required if wanting to turn on for columns
+    * Enables the use of html and js for the whole table, this is required if wanting to turn on for columns
```
alexcjohnson, line 104:
```suggestion
    * Allow strings containing JavaScript code or HTML in certain props.
    * If your app stores Dash layouts for later retrieval this is dangerous
    * because it can lead to cross-site-scripting attacks.
```

src/lib/components/AgGrid.react.js
alexcjohnson: OK I see now. This prop isn't _enabling_ update, it _is_ the update.

Can we then rename it to just `updateRows`, with a description something like:

> Provide new data to one or more existing rows. The rows to be replaced are the ones with matching row IDs, based on the field specified by `setRowId`.

And with that in mind actually, can we change the prop `setRowId` to `rowIdField` (including in the description above), with a description something like:

> The field within each row of `rowData` that will be used as the ID when identifying rows for updates.

BSd3v: Yes. We could do that. I was just using setRowId because it correlates to getRowId in ag-grid

alexcjohnson (approved):
OK! I still think we can get rid of a few JSON cycles, and we need tests but we'll do that after I get the framework set up in this repo.

Let's get rid of `parentState` as a prop of the outer component, just make it a separate prop of the fragment, then this is good to go! 💃

PR #9, opened 2023-01-25
allow html in stock demos

alexcjohnson:
@BSd3v I couldn't figure out how to get this working with `value(Formatter|Getter)Function`. This makes these stock demos work again but perhaps you can help me figure it out without `dangerously_allow_html`?


BSd3v:
@alexcjohnson, I adjusted the function in https://github.com/plotly/dash-ag-grid/pull/8 to follow this pattern for value parsing:

```
"valueGetter": {"function":"Round(data.price * data.quantity)"},
"valueFormatter": {"function":"toFixed(value)"},
```

The assets folder will also need to contain the js file

PR #19, opened 2023-02-02
add n_clicks to cellClicked 

### LINKED ISSUE #17: AgGrid cellClicked not working twice in a row
The event cellClicked will not be activated twice for the same cell in a row.

Here is and example:
```python
import pandas as pd
from dash import html,Input, Output, Dash
import dash_bootstrap_components as dbc
import dash_ag_grid

#
app = Dash(__name__, external_stylesheets=[dbc.themes.BOOTSTRAP])

# sample dataframe
df = pd.DataFrame({"col1": [1, 2, 3], "col2": ["A", "B", "C"]})

# column definition
column_defs = [
    {
        "headerName": "col1",
        "field": "col1",
    },
    {
        "headerName": "col2",
        "field": "col2",
    },
]
# layout
app.layout = html.Div(
    [
        dash_ag_grid.AgGrid(
            id="my-grid",
            columnDefs=column_defs,
        ),
    ]
)



@app.callback(
    Output("my-grid", "rowData"),
    Input("my-grid", "cellClicked")
)
def remove_line(cell_clicked):
    if cell_clicked is not None:
        col_id = cell_clicked["colId"]
        row_index = cell_clicked["rowIndex"]
        print(f"You clicked the icon in row {row_index} and col {col_id}")
    return df.to_dict("records")

if __name__ == '__main__':
    app.run_server(debug=True)
```

When you clicked twice on the same cel you get only one activation of the callback function:
```
You clicked the icon in row 0 and col col2
```
Instead you should have:
```
You clicked the icon in row 0 and col col2
You clicked the icon in row 0 and col col2
```

The only way to have two activation on the same cell is if you click on another and then go back:
```
You clicked the icon in row 1 and col col2
You clicked the icon in row 2 and col col2
You clicked the icon in row 1 and col col2
```

AnnMarieW:
fixes #17

Added n_clicks data so `setProps` would update the data when the same cell is clicked on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source D
   - Fix issue where columns would not take edits or adjustments due to becoming static
 
 - [#6](https://github.com/plotly/dash-ag-grid/pull/6) Fix props issue for `enableAddRows`
+- [#19](https://github.com/plotly/dash-ag-grid/pull/19) Fixed `cellClicked` as reported in [#17](https://github.com/plotly/dash-ag-grid/issues/17)
 
 ## [1.3.2] - 2023-01-13
 
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -35,8 +35,6 @@ export default class DashAgGrid extends Component {
             openGroups: new Set(),
             filterModel: {},
             dangerously_allow_html: JSON.parse(JSON.stringify(this.props.dangerously_allow_html)),
-            n_clicks: 0
-
         };
 
         if (this.props.rowData) {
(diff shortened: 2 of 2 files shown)


BSd3v:
@alexcjohnson, my thoughts are we should use a timestamp instead of n_clicks.
- n_clicks is an extra state, though small, to keep track of
- a timestamp may be useful as some point in time for gauging an user's level of interaction?

alexcjohnson:
Ooh yeah I like the `timestamp` idea. As we go along we might end up adding other timestamps as well, and there may be use cases where you can see which is the latest timestamp across different classes of actions, which wouldn't be possible with `n_clicks`. Also `n_clicks` itself is ambiguous, does it mean number of clicks on this cell? on any cell? for this event (ie 2 would be a doubleclick)? We could document but if we use `timestamp` there's no need.

alexcjohnson (approved):
💃 Excellent!

PR #28, opened 2023-02-09
Dangerously_allow_code and Function Parsing updates

BSd3v:
- allowed for other column prop functions to pass without disabling them if `dangerously_allow_code` is not passed
- copying over `columnDef` `dangerously_allow_code` to allow for the prop to be placed only on the grid level
- created a unified function parser to be easier to maintain
- brought the props to iterate through into a separate js `utils/functionVars.js` file
- switched to keeping `params` together instead of splitting into separate keys, to allow for easier transition to using AG Grid docs
- updated `test/test_add_remove_update_rows.py` to apply the new variable configuration


alexcjohnson (approved):
Nice! Need to resolve conflicts, and let's remove the [commented-out code block](https://github.com/plotly/dash-ag-grid/pull/28/files#diff-90827d2da7611ae1bb055beccd79eaa1e9a32025687fd333e0605d8e13eb53f1R642) - it's there in the git history if we want it back 😄 - then this is ready to go 💃

PR #33, opened 2023-02-13
Test framework

alexcjohnson:
To run these tests locally, just make sure you have dash installed for testing - you can run `pip install -r tests/requirements.txt` - then just `pytest` to run all the tests, or `pytest -k cd001` will just run the one test with `cd001` in its function name. `pytest -k cd001 --pause` will run that same single test, but drop into `pdb` right after opening the browser window, so you can debug the test both in python (`n` a few times and you'll be in the test code) and in the browser (open the js console and do whatever you like)

I added two example tests - one showing dragging & pinning columns and verifying that they're drawn as expected after, another filtering from the ui. I didn't add any tests with dash callbacks, but those work the same way. Both tests use the `utils.Grid` class to centralize all the ag-grid-specific selectors.


alexcjohnson:
I had to [downgrade to node 16](https://github.com/plotly/dash-ag-grid/pull/33/commits/ab5ea878032cd9359ecb602391e35e739258bf5b) - pretty sure this is fixed in recent versions of Webpack (see [this SO post](https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported)), but this project is still on v4 so I'll upgrade to v5 in a separate PR and see if that lets us upgrade to node 18

PR #47, opened 2023-02-22
Fixing `virtualRowData` 

BSd3v:
- by adding `rowModelType` as default `'clientSide'`
- added test for sort, and filter to update `virtualRowData`


alexcjohnson (approved):
💃 Love it!

PR #49, opened 2023-02-23
Cell and row styling including class rules

BSd3v:
- added support for `cellClassRules`, `rowClassRules` and `getRowClass`
- added tests to check for each class rule under `test_class_rules.py`
- adjusted test for conditional formatting to include row conditional formatting

### DIFF
--- a/tests/test_class_rules.py
+++ b/tests/test_class_rules.py
@@ -1,0 +1,80 @@
+import dash_ag_grid as dag
+from dash import Dash, html, dcc
+from . import utils
+import time
+from dash.testing.wait import until
+
+def test_cr001_class_rules(dash_duo):
+    app = Dash(__name__,
+                    meta_tags=[{'http-equiv': 'content-security-policy',
+                                'content': "default-src 'self'; script-src 'self' 'unsafe-inline';"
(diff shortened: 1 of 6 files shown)


tests/test_class_rules.py
```diff
@@ -0,0 +1,78 @@
...
+            dag.AgGrid(
+                id="grid",
+                columnDefs=columnDefs,
+                rowData=rowData,
+                defaultColDef={
+                    "editable": True,
+                },
+                rowClassRules={"orange": "params.data.make == 'Ford'",
+                               "red": "params.data.make == 'Porsche'",
+                               "blue": "params.data.make == 'Toyota'"},
```
alexcjohnson, line 44: For clarity, can we give the classes in `rowClassRules` different names & functions from the ones in `columnDefs`?

BSd3v: Sure, I'm adding `row-` in front of them.

BSd3v: Also made the functions query the model instead of the make

BSd3v: @alexcjohnson :)

alexcjohnson (approved):
💃 Great!

PR #62, opened 2023-03-02
feat: clear rows for grid in infinite row mode

fvdnabee:
Feedback welcome: e.g. a better name for the property.


BSd3v:
This was addressed by #81

PR #67, opened 2023-03-03
Function parsing recursive columndefs

BSd3v:
- adding more functions to be available for parsing
- allowed for recursively going through `columnDefs` -> `children` and master detail
- added various tests for functionality


tests/test_custom_row_selectable.py
```diff
@@ -0,0 +1,127 @@
+import dash_ag_grid as dag
```
AnnMarieW, line 1: How about making a simpler test? Maybe something like:

```

import dash_ag_grid as dag
from dash import Dash, html

def test_cr002_custom_row_selectable(dash_duo):

    app = Dash(__name__)

    columnDefs = [
        {    
            "field": "ticker",
            "headerCheckboxSelection": True,
            "checkboxSelection": True,
            "showDisabledCheckboxes": True,
        }
    ]
    grid = dag.AgGrid(
        id="grid",
        columnDefs=columnDefs,
        rowData=[{"ticker": "AAPL"}, {"ticker": "MSFT"}],
        dashGridOptions={'isRowSelectable': {'function': 'params.data.ticker == "AAPL" ? true: false'}},
        rowSelection="multiple"
    )

    app = Dash(__name__)

    app.layout = html.Div(grid)

    dash_duo.start_server(app)

    grid = utils.Grid(dash_duo, "grid")

    grid.wait_for_cell_text(0, 1, "Apple")

    assert dash_duo.find_element('#grid [row-index="0"] [aria-colindex="1"] '
                                 '.ag-selection-checkbox input').get_attribute('disabled') == None
    assert dash_duo.find_element('#grid [row-index="1"] [aria-colindex="1"] '
                                 '.ag-selection-checkbox input').get_attribute('disabled')


```

BSd3v: That’s fine for me.

AnnMarieW:
Also need to add to `functionVars.js`

 -  `predicate` function for use in custom filter options - https://www.ag-grid.com/react-data-grid/filter-provided-simple/#example-custom-filter-options

tests/test_custom_row_selectable.py
BSd3v: See https://github.com/plotly/dash-ag-grid/pull/67/commits/8bdd11d8ff2d513ad32c88467504aeb42d3f87a9

BSd3v:
> Also need to add to `functionVars.js`
> 
> * `predicate` function for use in custom filter options - https://www.ag-grid.com/react-data-grid/filter-provided-simple/#example-custom-filter-options
> * all the pagination functions as shown:  https://www.ag-grid.com/react-data-grid/row-pagination/#pagination-properties

Updated to allow for parsing of `filterParams` to get `valueGetter` function and also allow `filterParams` to be a variable/function in the namespace as well

alexcjohnson (approved):
💃 as discussed, I'll dig into nested detail grids in an upcoming PR to refactor how we secure these function props so that we don't push the updates back into the props. Otherwise this looks great! Nice job tracking down all these places AG Grid can `eval`!

PR #68, opened 2023-03-06
simplified custom components test

AnnMarieW:
Simplified the custom component test to removed the `yfinance` dependency and the `stock_data.csv` file


tests/test_custom_components.py
```diff
@@ -1,70 +1,24 @@
...
-import yfinance as yf
 from . import utils
 import json
-import os
-import time
+
 
 def test_cc001_custom_components(dash_duo):
 
-    app = Dash(__name__,
```
BSd3v, line 12: I would prefer leaving the meta tags in here, so that we make sure the components load even when the application is locked down.

tests/test_custom_components.py
```diff
@@ -1,70 +1,24 @@
...
-               meta_tags=[{'http-equiv': 'content-security-policy',
-                           'content': "default-src 'self'; script-src 'self' 'unsafe-inline';"
-                                      " style-src https://* 'self' 'unsafe-inline'; "
-                                      "font-src data: https://* 'self' 'unsafe-inline';"
-                                      "img-src data: https://* 'self'"}],
-               )
```
BSd3v, line 18: I would prefer leaving the meta tags in here, so that we make sure the components load even when the application is locked down.

AnnMarieW:
re-created in #125

PR #70, opened 2023-03-06
Click data cell renderer data

BSd3v:
- Changing `clickData` to `cellRendererData` to more closely line up with what this does

### DIFF
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -551,14 +551,18 @@ export default class DashAgGrid extends Component {
     }
 
     generateRenderer(Renderer) {
-        const {setProps, dangerously_allow_code, cellRendererData} = this.props;
-
-        const setCellProps = (cellRendererData) => {
-            setProps({cellRendererData, cellRendererData_timestamp: Date.now()});
-        };
+        const {setProps, dangerously_allow_code} = this.props;
 
         return (props) => (
-            <Renderer setData={setCellProps} dangerously_allow_code={dangerously_allow_code} {...props}></Renderer>
+            <Renderer setData={(value) => {
+                setProps({cellRendererData: {
+                value,
+                colId: props.column.colId,
+                rowIndex: props.rowIndex,
+                rowId: props.node.id,
+                timestamp: Date.now()}
+                });
+            }} dangerously_allow_code={dangerously_allow_code} {...props}></Renderer>
         );
     }
 
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -556,7 +556,7 @@ export default class DashAgGrid extends Component {
         const {setProps, dangerously_allow_code} = this.props;
 
         const setCellProps = (props) => {
-            setProps({clickData: props.clickData, hoverData: props.hoverData});
+            setProps({cellRendererData: props.cellRendererData});
         };
 
         return (props) => (
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -551,14 +551,14 @@ export default class DashAgGrid extends Component {
     }
 
     generateRenderer(Renderer) {
-        const {setProps, dangerously_allow_code} = this.props;
+        const {setProps, dangerously_allow_code, cellRendererData} = this.props;
 
-        const setCellProps = (props) => {
-            setProps({cellRendererData: props.cellRendererData});
+        const setCellProps = (cellRendererData) => {
+            setProps({cellRendererData, cellRendererData_timestamp: Date.now()});
         };
 
         return (props) => (
-            <Renderer setProps={setCellProps} dangerously_allow_code={dangerously_allow_code} {...props}></Renderer>
+            <Renderer setData={setCellProps} dangerously_allow_code={dangerously_allow_code} {...props}></Renderer>
         );
     }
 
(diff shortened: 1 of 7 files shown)


tests/assets/dashAgGridComponentFunctions.js
```diff
@@ -113,4 +113,28 @@ dagcomponentfuncs.myCustomButton = function (props) {
...
+
+    if (!props.value) {
+        return React.createElement('button')
+    }
+
+    function onClick() {
+        setProps({cellRendererData: 'updated'})
+    }
+
+    const id = JSON.stringify({'index': props.rowIndex, 'type':'customButton'})
```
AnnMarieW, line 130: Might be a performance issue with lots of components?

BSd3v: It’s just really an example of an id, the id isn’t actually necessary, it just helps react to stop complaining. 

Or were you talking about the setProps?

src/lib/fragments/AgGrid.react.js
```diff
@@ -555,7 +554,7 @@ export default class DashAgGrid extends Component {
         const {setProps, dangerously_allow_code} = this.props;
 
         const setCellProps = (props) => {
-            setProps({clickData: props.clickData, hoverData: props.hoverData});
+            setProps({cellRendererData: props.cellRendererData});
```
alexcjohnson, line 557: 💄 
```js
const setCellProps = ({cellRendererData}) => {
    setProps({cellRendererData});
}
```

Makes me wonder though - now that it's clear that this is only going to a single prop, should we pass it to renderers as a `setData` prop rather than `setProps`, ie in [rowMenuRenderer](https://github.com/plotly/dash-ag-grid/pull/70/files#diff-4aa177d562cc6efb214ba041c20342ed5bb67d7a994c0d7f8d8d00a5365e612fR25) instead of `setProps({cellRendererData: {value, data}})` you would just call `setData({value, data})`, then the function here would just be:

```js
const setCellProps = (cellRendererData) => {
    setProps({cellRendererData});
}
```

But then I also wonder if this is enough for real use cases... would we want to add info about which cell this is, as well as `timestamp` like we have on `clickData`?

BSd3v: As far as that info, it’s up to the dev to decide what they want to use or provide. 

Unless we automatically tack on a timestamp? Like cellRendererData_timestamp: Date.now()?

BSd3v: See: https://github.com/plotly/dash-ag-grid/pull/70/commits/6f07a05ae567dce907f4633ea8cba903d8959d1f

alexcjohnson: Sorry, I misspoke, I meant to refer to `onCellClicked`, where we call:
```js
setProps({cellClicked: {value, colId, rowIndex, timestamp}})
```
`colId` and `rowIndex` give you an exact pointer to the cell that was clicked (would be even more useful if we add `rowId` to this, at least when `getRowId` is defined), `timestamp` isn't generally intended to really be used, just ensures uniqueness, and `value` is the cell data you clicked on.

So I feel like we should follow that same pattern here. When defining a custom renderer, you just call `setData(arbitraryObject)` but you shouldn't need to explicitly include the extra info needed to identify the cell and make the object unique. we should take your `arbitraryObject` and set it as the `value` inside `cellRendererData`, but also include `colId, rowIndex, rowId, timestamp`.

Look at the row menu renderer example https://dashaggrid.pythonanywhere.com/components/row-menu: The renderer chooses to include the full row data, which if they're unique allows you to identify which row you're in. If they aren't unique you can't identify the row, and it does nothing to tell you which column the event came from so if you had these menus in multiple columns with matching option labels you'd need to build uniqueness into the `value` field. If the table has lots of columns, providing the full row is also more info than you generally need. I'd love if `rowMenuRenderer` could just call `setData(value)` and all the other pieces (`colId, rowIndex, rowId, timestamp`) are added automatically. That would make any additional renderers people make more powerful without needing to think about any of this.

AFAICT if we move `setCellProp` into the function `return (props) => ( <Renderer ... /> )` we have all the info we need to do this? And then of course with `timestamp` built into the `cellRendererData` prop we don't need the `cellRendererData_timestamp` prop anymore.

BSd3v: see https://github.com/plotly/dash-ag-grid/pull/70/commits/f662c2ee71142a5fb901eb570f7d65522e5502e6

src/lib/renderers/rowMenuRenderer.js
```diff
@@ -20,14 +20,12 @@ export default function RowMenuRenderer(props) {
                 target: {value},
             } = e;
 
-            if (setProps && value) {
-                setProps({
-                    clickData: {value, data},
-                });
+            if (setData && value) {
+                setData({value, data});
```
alexcjohnson, line 24: Can we just do `setData(value)` now that `cellRendererData` contains all the other fields? Then in test `rm001` we can use `colId`, `rowId`, and `rowIndex` in the output since we know what they should be, plus a simple `assert 'timestamp' in data` (yes, that assertion can go inside the callback because if it fails the callback won't return anything and our other assertions later will fail).

alexcjohnson (approved):
💃 excellent!

alexcjohnson:
As @BSd3v and I discussed, there's a slight loss of functionality by removing `data` from the `rowMenuRenderer` event data. The rationale is that we've more than made up for that by adding `rowIndex`, `rowId`, and `colId`, and `data` is potentially a lot of bytes and mostly not necessary. Now is the time for breaking changes so we removed it. But if anyone complains we can always add it back in later, and if we did that we'd add it to all `cellRendererData` values, not just `rowMenuRenderer`, by adding it to the event handler defined in `generateRenderer`.

PR #74, opened 2023-03-07
Dep updates 20230307

alexcjohnson:
- Upgrade to AG Grid 29.1, and switch `>=` to `^` so we don't accidentally get v30 whenever that's released
- Minor lint & build updates
- Add `eslint` linting to CI - but for now leaving `prettier` out of that, there will be lots of changes when we do that so I'll wait until there is less other work in progress.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,9 +23,12 @@ jobs:
       - run:
           name: Build package
           command: |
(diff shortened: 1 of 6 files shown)


ndrezn:
I am excited for the `prettier` PR 💄

PR #76, opened 2023-03-08
Adding logging function

BSd3v:
- adding logging function available by default, available via `{"function": "log()"}`

### DIFF
--- a/src/lib/renderers/customFunctions.js
+++ b/src/lib/renderers/customFunctions.js
@@ -2,6 +2,6 @@ export const customFunctions = {
     "Math": Math,
     "Number": Number,
     "log": function () {
-        console.log('you are logging from an AG grid function');
+        console.log('you are logging from an AG grid function', ...arguments);
     }
 }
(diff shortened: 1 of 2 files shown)


BSd3v:
When utilized, you can click on the link in the console log and set a breakpoint, then upon next run you can view the variables by selecting `value` from the `Call Stack`

image

src/lib/renderers/customFunctions.js
```diff
@@ -1,4 +1,7 @@
 export const customFunctions = {
     "Math": Math,
-    "Number": Number
+    "Number": Number,
+    "log": function () {
+        console.log('you are logging from an AG grid function');
```
alexcjohnson, line 5:
```suggestion
        console.log('you are logging from an AG grid function', ...arguments);
```
So if the user wants to call `log(1,2,3)` or whatever, they can do that, and they'll see the values directly, without setting a breakpoint.

alexcjohnson (approved):
💃

PR #92, opened 2023-03-16
Fixing `defaultStyle` in the `cellStyle` prop when only listed

BSd3v:
- added test for `defaultStyle` in the `cellStyle` when `styleConditions` is not also provided
- added tests for `defaultStyle`


alexcjohnson (approved):
💃

PR #102, opened 2023-03-18
Refactor prop cleaning

alexcjohnson:
Change from in-place cleaning of function props to building new prop objects with function strings converted to functions.

@BSd3v please take a look - a few tests are failing and I haven't looked at why yet, and I'm still not 100% sure I understand all the props (which may be why some tests are failing) but I think you'll get the point of the refactoring I'm trying to do, so that the props provided by the user are kept unchanged even while the props seen by AG Grid are converted and sanitized.

I did a few things to improve performance as well: splitting out `esprima.parse` from the execution of the function, which only needs `evaluate`; turning the `functionVars` categories into maps instead of arrays so we can do a hash lookup instead of `Array.includes`...

### DIFF
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -2,16 +2,21 @@ import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 import * as evaluate from 'static-eval';
 import * as esprima from 'esprima';
-import {omit, equals, isEmpty} from 'ramda';
+import {equals, has, isEmpty, map, mapObjIndexed, omit} from 'ramda';
 import {
     propTypes as _propTypes,
     defaultProps as _defaultProps,
 } from '../components/AgGrid.react';
 import {
-    expressWarn,
-    gridFunctions,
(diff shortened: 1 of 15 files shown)


src/lib/fragments/AgGrid.react.js
```diff
@@ -668,65 +584,51 @@ export default class DashAgGrid extends Component {
...
-        return defaultStyle ? defaultStyle : null;
-    }
+    parseFunctionNoParams = memoizeWith(String, (funcString) => {
+        const parsedCondition = esprima.parse(funcString).body[0].expression;
+        const context = {
+            d3,
+            ...customFunctions,
+            ...window.dashAgGridFunctions,
+        };
+        return evaluate(parsedCondition, {...context});
```
alexcjohnson, line 604:
```suggestion
        return evaluate(parsedCondition, context);
```

BSd3v:
I need to add a couple of tests, one for infinite scroll, and then other for the recursive functions to query the data from the server vs directly from the grid.

Issue #109, opened 2023-03-21
Feature Request -- return filter query 


alisonmoss3:
Hello! 

Thank you so much for making this package public, it is AMAZING!! 

I did have one request I was hoping could be added. 

I'm doing a lot of work with filtering tables and storing the results alongside the query, and I see that you can pull out the resulting data using virtualRowData but I was wondering if there is any way to access the filter query on that table from a callback. 

For example, if I start with a table of 100 rows and filtering on 'column A > 2' results in 50 rows, I can access those 50 rows of data but I can't see a way to retrieve 'column A > 2'. With more complex filters on multiple columns this would be extremely helpful. 

Thank you!

BSd3v:
Hello @alisonmoss3,

Yes, this is often requested and can be very helpful. We will add it before the full release of 2.0.

alisonmoss3:
Thank you! Looking forward to it!

Issue #99, opened 2023-03-18
autoSizeAllColumns bug


AnnMarieW:
Error when updating  `autoSizeAllColumns` prop in  a callback:

![image](https://user-images.githubusercontent.com/72614349/226110756-6d6845af-53a2-4fde-b098-1368faade52e.png)

Here is MRE
```python

from dash import Dash, html, Input, Output
import dash_ag_grid as dag
import pandas as pd

df = pd.read_csv("https://raw.githubusercontent.com/plotly/datasets/master/gapminder2007.csv")

app = Dash(__name__)

grid = dag.AgGrid(
    id="size-grid",
    rowData=df.to_dict("records"),
    columnDefs=[{"field": i, "id": i} for i in df.columns],
    defaultColDef={"resizable": True},
    columnSize="sizeToFit",
)


app.layout = html.Div([
    html.Button("Auto Size", id="auto-size"),
    grid
])


@app.callback(
    Output("size-grid", "autoSizeAllColumns"),
    Input("auto-size", "n_clicks"),
   # prevent_initial_call=True
)
def change_size(_):
    return True


if __name__ == "__main__":
    app.run_server(debug=True)


```

BSd3v:
This isn’t a bug, the grid isn’t ready, that is why it can’t be triggered immediately. 

This needs to be delayed until the grid is ready if provided.

AnnMarieW:
True - if you un-comment  the `prevent_initial_call=True` in the example, the error goes away.  However, it's not always possible to prevent callbacks from firing in a multi-page apps.  Can the this error be handled in the component?  https://github.com/plotly/dash-ag-grid/blob/dev/src/lib/fragments/AgGrid.react.js#L844

BSd3v:
That’s what my previous secondary statement was about. 

This should be true for all event triggers.

PR #111, opened 2023-03-24
adding support for `filterModel`

BSd3v:
- info from the grid (#109)
- added output for `test_filter`
- added check for action props to delay until the grid is ready (#99)
- adjusted and fixed html templates (#116)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,9 @@ Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source D
   - allowed for functions, styleConditions and regular dictionaries to be passed to the `cellStyle` on all levels
   - added `rowId` to `cellClicked` data
 
+- [#111](https://github.com/plotly/dash-ag-grid/pull/111)
+  - fixing templates to only populate when `dangerously_allow_code=True`
+
 ### Fixed
 - [Overhaul commit](https://github.com/plotly/dash-ag-grid/commit/b888d6ab4fcb4afac187492e8b6c9cf0d0f8842b)
   - Fix issue where conditional formatting was not applied to nested columns
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,10 @@ Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source D
 - [#81](https://github.com/plotly/dash-ag-grid/pull/81)
   - added tests for `cellClicked` data
 
+- [#111](https://github.com/plotly/dash-ag-grid/pull/111)
+  - added `filterModel` prop in order to capture the grid's active filters
+
+
 ### Updated
 - [Overhaul commit](https://github.com/plotly/dash-ag-grid/commit/b888d6ab4fcb4afac187492e8b6c9cf0d0f8842b)
   - Update AG Grid from v27.x to v29.x - see [AG Grid Changelog](https://www.ag-grid.com/changelog/) for details.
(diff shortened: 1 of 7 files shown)


src/lib/fragments/AgGrid.react.js
```diff
@@ -329,16 +329,14 @@ export default class DashAgGrid extends Component {
...
             this.state.gridApi.forEachNodeAfterFilter((node) => {
                 virtualRowData.push(node.data);
             });
-
-            const filterModel = this.state.gridApi.getFilterModel();
-            this.setState({filterModel});
-            setProps({virtualRowData});
         }
+        const filterModel = this.state.gridApi.getFilterModel();
+        setProps({virtualRowData, filterModel});
```
alexcjohnson, line 339: Do we actually want to be setting `virtualRowData` if `rowModelType` is not `'clientSide'`? If not, we can make a `const propsToSet = {filterModel}` and inside the `if` we set `propsToSet.virtualRowData = virtualRowData`.

BSd3v: Ok, yes. That's a better idea. 

I was avoiding the setProps for the virtualRowData by itself for obvious reasons, but also, it was clearing the filters too.

src/lib/fragments/AgGrid.react.js
```diff
@@ -528,6 +539,43 @@ export default class DashAgGrid extends Component {
...
+            this.deleteSelectedRows(false);
+        }
+
+        setProps({
+            exportDataAsCsv: false,
+            selectAll: false,
+            deselectAll: false,
+            autoSizeAllColumns: false,
+            deleteSelectedRows: false,
+        });
```
alexcjohnson, line 577: Also this one - we should only `setProps` the ones that weren't already `false`.

BSd3v: Alright, I'm using the same way the above works. `propsToSet` then build through each, then if `propsToSet` not empty, it `setProps`

CHANGELOG.md
```diff
@@ -76,6 +76,9 @@ Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source D
   - allowed for functions, styleConditions and regular dictionaries to be passed to the `cellStyle` on all levels
   - added `rowId` to `cellClicked` data
 
+- [#111](https://github.com/plotly/dash-ag-grid/pull/111)
+  - fixing templates to only populate when `dangerously_allow_code=True`
```
alexcjohnson, line 80: That note belongs under `Fixed` - but we should also note the addition of `filterModel` in the `Added` section

src/lib/fragments/AgGrid.react.js
```diff
@@ -528,6 +548,47 @@ export default class DashAgGrid extends Component {
...
+
+        if (deleteSelectedRows) {
+            this.deleteSelectedRows(false);
+            propsToSet.deleteSelectedRows = false;
+        }
+
+        if (!isEmpty(propsToSet)) {
+            setProps({
+                ...propsToSet,
+            });
```
alexcjohnson, line 589:
```suggestion
            setProps(propsToSet);
```

alexcjohnson (approved):
Excellent! 💃 pending changelog edit.

CHANGELOG.md
BSd3v: adjusted via https://github.com/plotly/dash-ag-grid/pull/111/commits/679ed74b1af6dbeff7c142d63c62c7f845e893c7

PR #124, opened 2023-04-01
add sizeColumnsToFit

### LINKED ISSUE #100: Feature Request add `sizeColumnsToFit` prop
Currently when `columnSize="sizeToFit" ` is set and you change the screen size, the only way to redo the `sizeToFit` is to refresh the screen. 

It would be great to have a `sizeColumnsToFit` prop so it's possible to call this AG Grid function from a a Dash callback. It would make it possible to have a "size to fit" button.

AnnMarieW:
Closes #100


AnnMarieW:
Closed by better solution in #132 :tada:

PR #129, opened 2023-04-02
updated docstring for suppressDragLeaveHidesColumns

### LINKED ISSUE #128: Misleadin docsting for `suppressDragLeaveHidesColumns` argument of AgGrid class in v2.0.0a5
What is default value?

```
- suppressDragLeaveHidesColumns (boolean; default True):
    If True, when you drag a column out of the grid (e.g. to the group
    zone) the column is not hidden. Default Value: False.
```

AnnMarieW:
Closes #128


alexcjohnson (approved):
💃 Good call - general rule, the only time a default should be described in the docstring is when it's dynamic, ie we don't specify it in `defaultProps` but the default behavior changes depending on some other prop values.

PR #130, opened 2023-04-03
Hot fix for applying `filterModel` initially

BSd3v:


src/lib/fragments/AgGrid.react.js
```diff
@@ -465,7 +465,9 @@ export default class DashAgGrid extends Component {
             });
         }
         if (!isEmpty(filterModel)) {
-            this.state.gridApi.setFilterModel(filterModel);
+            if (this.state.gridApi) {
+                this.state.gridApi.setFilterModel(filterModel);
```
alexcjohnson, line 469: Could we simplify `onRowDataUpdated` by (1) pulling out `gridApi` with `const {openGroups} = this.state`, (2) bail out right at the beginning if `gridApi` is falsy? Looks like everything here will fail if the grid isn't ready, right?

alexcjohnson (approved):
💃

PR #132, opened 2023-04-05
Column size updates

BSd3v:
- removed `autoSizeAllColumns`
- made `columnSize` act upon update
- added new `columnSize` available, `responsiveSizeToFit` which will adjust column sizes based upon grid size and columns added or removed
- updated sizing tests


src/lib/components/AgGrid.react.js
```diff
@@ -288,11 +275,25 @@ DashAgGrid.propTypes = {
...
+        null,
+    ]),
+
+    /**
+     * column sizing options as derived by the columnSize choice,
+     * autoSize https://www.ag-grid.com/react-data-grid/column-sizing/#autosize-column-api,
+     * keys are provided, it is assumed that only these columns will be auto-sized,
+     * sizeToFit/responsiveSizeToFit https://www.ag-grid.com/react-data-grid/column-sizing/#size-columns-to-fit
+     */
+    columnSizeOptions: PropTypes.object,
```
alexcjohnson, line 296: Can we give this more precision?

```suggestion
    /**
     * Options to customize the columnSize operation.
     * autoSize calls either autoSizeColumns or autoSizeAllColumns, see:
     * https://www.ag-grid.com/react-data-grid/column-sizing/#autosize-column-api,
     * and sizeToFit and responsiveSizeToFit call sizeColumnsToFit, see:
     * https://www.ag-grid.com/react-data-grid/column-sizing/#size-columns-to-fit
     */
    columnSizeOptions: PropTypes.exact({
        /**
         * for (responsive)sizeToFit: per-column minimum and maximum width, in pixels.
         */
        columnLimits: PropTypes.arrayOf(
            PropTypes.exact({
                key: PropTypes.string,
                minWidth: PropTypes.number,
                maxWidth: PropTypes.number,
            })
        ),
        /**
         * for (responsive)sizeToFit: default minimum width, in pixels, if not overridden by columnLimits
         */
        defaultMinWidth: PropTypes.number,
        /**
         * for (responsive)sizeToFit: default maximum width, in pixels, if not overridden by columnLimits
         */
        defaultMaxWidth: PropTypes.number,
        /**
         * for autoSize: list of column keys to autosize. If omitted, all columns will be autosized.
         */
        keys: PropTypes.arrayOf(PropTypes.string),
        /**
         * for autoSize: should we skip header contents and only consider cell contents?
         */
        skipHeader: PropTypes.bool,
    }),
```

alexcjohnson (approved):
💃 Looks great - just a suggestion to make the prop type more precise, otherwise I think this is good to go. I still have some questions about how this will play with saving & restoring a grid, especially if the data may be different in the restored view. But I think in any event we'll be able to use this as a starting point and make non-breaking changes later if we find that something else is needed for that use case.

PR #145, opened 2023-04-17
Adding support for AlignedGrids and pagination controls

BSd3v:
- adding support for `alignedGrids`
- adjusted test for dragging to link two grids and maintain linking through changing columns
- added ability for functions with `tooltipComponentParams`
- added `paginationInfo` for read-only info from the grid's pagination
- added `paginationGoTo` to navigate to different pages


src/lib/fragments/AgGrid.react.js
```diff
@@ -317,6 +355,14 @@ export default class DashAgGrid extends Component {
             if (GRID_MAYBE_FUNCTIONS_NO_PARAMS[target]) {
                 return this.convertMaybeFunctionNoParams(value);
             }
+            if (target === 'alignedGrids') {
+                return (
+                    document.getElementById(value) && [
+                        findReact(document.getElementById(value)).reference
+                            .current,
```
alexcjohnson, line 362: Isn't this ref just a DOM element? And if so, can we find it directly rather than going through React internals? Maybe it's just this?
```suggestion
                        document.getElementById(value).childNodes[0],
```

Also, the underlying API is an array, can use an array of IDs and map each one to the appropriate element? Maybe that just works for multiple aligned grids?

BSd3v: No, its not mapped to the DOM, its linked on the React component. Changes to the grid makes changes to the API of the connected grid.

Yes, we could map it to an array, wasnt sure if we wanted to tackle this just yet.

src/lib/fragments/AgGrid.react.js
```diff
@@ -126,6 +129,25 @@ export default class DashAgGrid extends Component {
...
+        if (gridApi) {
+            setProps({
+                paginationInfo: {
+                    paginationIsLastPageFound:
+                        gridApi.paginationIsLastPageFound(),
+                    paginationGetPageSize: gridApi.paginationGetPageSize(),
+                    paginationGetCurrentPage:
+                        gridApi.paginationGetCurrentPage(),
+                    paginationGetTotalPages: gridApi.paginationGetTotalPages(),
+                    paginationGetRowCount: gridApi.paginationGetRowCount(),
```
alexcjohnson, line 147: Rather than matching the method names exactly, let's use the obvious names you'd give the results of these methods - ie strip off `paginationGet` so each is a noun. And then so it's fully self-documenting, let's describe them in `propTypes`, ie `paginationInfo: PropTypes.exact({...})`
```suggestion
                    isLastPageFound: gridApi.paginationIsLastPageFound(),
                    pageSize: gridApi.paginationGetPageSize(),
                    currentPage: gridApi.paginationGetCurrentPage(),
                    totalPages: gridApi.paginationGetTotalPages(),
                    rowCount: gridApi.paginationGetRowCount(),
```

src/lib/fragments/AgGrid.react.js
```diff
@@ -386,9 +408,20 @@ export default class DashAgGrid extends Component {
     }
 
     componentDidMount() {
+        const {id} = this.props;
+        if (id) {
+            agGridRefs[id] = this.reference.current;
```
alexcjohnson, line 413: What if someone changes the ID? I'm not sure if the renderer would actually do this... might always unmount and create a new component. But in case it did, we should probably have `componentDidUpdate` check for a changed ID, delete the old ref and add the new one.

src/lib/fragments/AgGrid.react.js
```diff
@@ -949,9 +1025,29 @@ export default class DashAgGrid extends Component {
...
+        let alignedGrids;
+        if (dashGridOptions) {
+            if ('alignedGrids' in dashGridOptions) {
+                alignedGrids = [];
+                const addGrid = (id) => {
+                    if (!agGridRefs[id]) {
+                        agGridRefs[id] = {api: null};
+                    }
+                    alignedGrids.push(agGridRefs[id]);
+                };
```
alexcjohnson, line 1035: There's a complication here with pattern-matching IDs - `props.id` will always be a string, because we handle that separately and stringify those objects when passing them from dash renderer to the actual components. But the renderer doesn't know to do that inside `alignedGrids` so we'll need to do it ourselves. And it's not just `JSON.stringify` either... we should probably export this from the renderer so it can be reused, but we don't so we'll need to copy it here. [Here's the source.](https://github.com/plotly/dash/blob/5fd9a14a05c46f5c15802109c263423930297cc8/dash/dash-renderer/src/actions/dependencies.js#L114-L123)

alexcjohnson: Also shouldn't we do this conversion inside `convertOne`? As is it'll also be inside of `convertedProps` so it's in principle ambiguous, right?

BSd3v: We cant memoize, this actually caused issues because even though the background object changed, the end result wasnt correct.

I removed it from `convertedProps`. 😄

src/lib/fragments/AgGrid.react.js
```diff
@@ -386,9 +417,20 @@ export default class DashAgGrid extends Component {
     }
 
     componentDidMount() {
+        const {id} = this.props;
+        if (id) {
+            agGridRefs[stringifyId(id)] = this.reference.current;
```
alexcjohnson, line 422: `this.props.id` is guaranteed to be a string already; so this one, and the ones in `componentWillUnmount` and `componentDidUpdate` don't need `stringifyId`, it's only reading `alignedGrids` where we need this.

src/lib/fragments/AgGrid.react.js
```diff
@@ -66,6 +69,39 @@ const xssMessage = (context) => {
...
+    on(id, data, callback) {
+        if (!(id in eventBus.listeners)) {
+            eventBus.listeners[id] = {};
+        }
+        eventBus.listeners[id][data] = callback;
+    },
+    dispatch(data) {
+        for (const id in eventBus.listeners) {
+            if (data in eventBus.listeners[id]) {
+                eventBus.listeners[id][data]();
```
alexcjohnson, line 85: To make it clearer what's going on here (`id` is listening to events from `data` which is also an ID), let's replace `data` with `targetId`.
```suggestion
    on(id, targetId, callback) {
        if (!(id in eventBus.listeners)) {
            eventBus.listeners[id] = {};
        }
        eventBus.listeners[id][targetId] = callback;
    },
    dispatch(targetId) {
        for (const id in eventBus.listeners) {
            if (data in eventBus.listeners[id]) {
                eventBus.listeners[id][targetId]();
```

src/lib/fragments/AgGrid.react.js
```diff
@@ -949,19 +1073,63 @@ export default class DashAgGrid extends Component {
...
+            if ('alignedGrids' in dashGridOptions) {
+                alignedGrids = [];
+                const addGrid = (id) => {
+                    const strId = stringifyId(id);
+                    eventBus.on(this.props.id, strId, () => {
+                        this.setState({
+                            toggle: this.state.toggle
+                                ? !this.state.toggle
+                                : false,
+                        });
```
alexcjohnson, line 1087: Isn't this always going to be `false`? But anyway, if you DO want to modify state based on its previous data, you're supposed to use the functional form `prevState=>updates` since in theory React can batch these updates asynchronously.
```suggestion
                        this.setState((prevState) => {
                            toggle: !prevState.toggle
                        });
```
That said, if the purpose here is to trigger a rerender so the alignment will update, it might be better (a) name it to make that clear (`state.rerender`?), (b) make it an incrementing counter (`this.setState((prevState) => {rerender: prevState.rerender + 1})`) in case there are two back-to-back events that toggle on and back off and make it look like nothing happened.

tests/test_column_drag.py
```diff
@@ -48,3 +49,138 @@ def test_cd001_drag_columns(dash_duo):
...
+    @app.callback(Output('hiding', 'children'), Input('link', 'n_clicks'),
+                  State('topGrid', 'columnState'))
+    def addGrid(n, s):
+        grid2 = AgGrid(
+            id="middleGrid",
+            columnDefs=columnDefs,
+            rowData=df.to_dict("records"),
+            columnSize="autoSize",
+            defaultColDef=defaultColDef,
+            dashGridOptions={'alignedGrids': 'bottomGrid'},
```
alexcjohnson, line 122: Weird that this is necessary, but the [AG Grid docs](https://www.ag-grid.com/javascript-data-grid/aligned-grids/#event-propagation) do seem to say this is how you need to do it. In principle it should be sufficient that you've already said the top grid is aligned to both middle and bottom, so implicitly the middle and bottom are also aligned to each other, and then we should be able to let AG Grid ensure all the right events are fired without infinite loops. At any rate that's how plotly.js does the equivalent thing, when you set several axes to have matching or scaled axes... you aren't even allowed to overspecify it.

src/lib/components/AgGrid.react.js
```diff
@@ -13,12 +13,7 @@ export default class DashAgGrid extends Component {
     constructor(props) {
         super(props);
 
-        this.state = {
-            gridApi: null,
-            columnApi: null,
-            openGroups: new Set(),
-            filterModel: {},
-        };
+        this.state = {mouted: false};
```
alexcjohnson, line 16:
```suggestion
        this.state = {
            mounted: false,
            rowTransaction: null,
        };
```

src/lib/fragments/AgGrid.react.js
```diff
@@ -381,12 +442,74 @@ export default class DashAgGrid extends Component {
...
+            return true;
+        }
+        if (gridApi) {
+            if (columnState) {
+                if (
+                    !equals(
+                        columnState,
+                        JSON.parse(
+                            JSON.stringify(gridColumnApi.getColumnState())
+                        )
```
alexcjohnson, line 494: I can't see the JSON cycle being necessary here, is it?
```suggestion
                        gridColumnApi.getColumnState()
```

alexcjohnson (approved):
💃 Looks good! Let's upgrade to the brand new AG Grid 29.3.2, and mention this in the changelog, then we're ready to go!

PR #155, opened 2023-04-24
bump to react-18

BSd3v:
- fix of `openGroups`
- fix `paginationGoTo`
- fix `rowMenuRenderer` `material-ui` to most recent version


src/lib/fragments/AgGrid.react.js
```diff
@@ -648,93 +732,16 @@ export default class DashAgGrid extends Component {
...
-        } = this.props;
+        const {rowModelType} = this.props;
 
-        const propsToSet = {};
         if (rowModelType === 'infinite') {
             params.api.setDatasource(this.getDatasource());
         }
 
-        this.setState({
+        this.setState(() => ({
```
alexcjohnson, line 741: Is it necessary to make this a function?

BSd3v: No, was trying to solve the bad `setState` issue, we can revert.

package.json
```diff
@@ -28,8 +28,10 @@
     "author": "Plotly <chris@plot.ly>",
     "license": "MIT",
     "dependencies": {
-        "@material-ui/core": "^4.11.4",
-        "@material-ui/icons": "^4.11.3",
+        "@mui/material": "^5.12.1",
+        "@mui/icons-material": "^5.11.16",
+        "@emotion/styled": "^11.10.6",
+        "@emotion/react": "^11.10.6",
```
alexcjohnson, line 34: Why are these `@emotion` packages here?

BSd3v: Dependency of `@mui/material`

alexcjohnson: I see, @BSd3v pointed me to this odd aspect of their [installation instructions](https://mui.com/material-ui/getting-started/installation/) that you need to explicitly install those even if you don't use them directly 🙄

AnnMarieW: Is there any way to just include the row menu component rather than the entire package?

PR #161, opened 2023-04-26
default style adjustment

BSd3v:
adjusted default style to be applied to the div as hardcoded instead of in the default props

### DIFF
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -1164,7 +1164,11 @@ export default class DashAgGrid extends Component {
         }
 
         return (
-            <div id={id} className={className} style={style}>
+            <div
+                id={id}
+                className={className}
+                style={{...{height: '400px', width: '100%'}, ...style}}
+            >
                 <AgGridReact
                     ref={this.reference}
                     alignedGrids={alignedGrids}
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -1167,7 +1167,7 @@ export default class DashAgGrid extends Component {
             <div
                 id={id}
                 className={className}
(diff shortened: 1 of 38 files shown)


alexcjohnson:
Ah interesting - yes, this makes sense. Let's just add a test to ensure that both `height` and `width` can be overridden, separately or together, via the `style` prop. Unless we already have one that I'm not seeing?

AnnMarieW:
Yes - based on this issue:  https://community.plotly.com/t/ag-grid-with-multi-page-apps/74728/15?u=annmariew

src/lib/fragments/AgGrid.react.js
```diff
@@ -1164,7 +1164,11 @@ export default class DashAgGrid extends Component {
         }
 
         return (
-            <div id={id} className={className} style={style}>
+            <div
+                id={id}
+                className={className}
+                style={{...{height: '400px', width: '100%'}, ...style}}
```
alexcjohnson, line 1170:
```suggestion
                style={{height: '400px', width: '100%', ...style}}
```

alexcjohnson (approved):
💃

mathman79:
Does this work correctly with ```dashGridOptions={"domLayout": "autoHeight"}```? I've been explicitly omitting `"height"` in the `style` because of this, but since moving from 2.0.0rc2 to 2.0.0 my tables look weird.

BSd3v:
This should be working.

Do you have an MRE to test?

AnnMarieW:
@mathman79 You can find an example of it working with 2.0.0 in the docs here:  https://dashaggrid.pythonanywhere.com/layout/grid-size
and as @BSd3v  mentioned a minimal example that shows the issue would help.

mathman79:
Ah, thanks a lot. I have to add ```{"height": "100%"}``` now to override the default. Thanks for the quick response.

PR #164, opened 2023-04-26
Selected rows changes

BSd3v:


src/lib/fragments/AgGrid.react.js
```diff
@@ -192,15 +194,62 @@ export default class DashAgGrid extends Component {
...
+                    gridApi.deselectAll();
+                } else {
+                    // keeps grid from rendering display unnecessarily
+                    gridApi.deselectAll();
+                    if (getRowId) {
+                        const parsedCondition = esprima.parse(
+                            getRowId.replaceAll('params.data.', '')
+                        ).body[0].expression;
+                        const mapId = {};
+                        selection.map((params) => {
```
alexcjohnson, line 235:
```suggestion
                        selection.forEach((params) => {
```
`map` works, but since we aren't using the return value it's best to use `forEach`.

src/lib/fragments/AgGrid.react.js
```diff
@@ -192,15 +194,62 @@ export default class DashAgGrid extends Component {
...
+                };
+                gridApi.forEachNode(test);
+            } else if (Object.keys(selection).includes('ids')) {
+                // keeps grid from rendering display unnecessarily
                 gridApi.deselectAll();
-            } else {
                 gridApi.forEachNode((node) => {
-                    const isSelected = selection.some(equals(node.data));
-                    node.setSelected(isSelected);
+                    if (selection.ids[node.id]) {
```
alexcjohnson, line 220: The ID map form is great for performance but not so user-friendly, let's have `selection.ids` be a list, and do the same conversion you do below to turn it into an ID map.

src/lib/components/AgGrid.react.js
```diff
@@ -604,7 +604,10 @@ DashAgGrid.propTypes = {
     /**
      * The actively selected rows from the grid (may include filtered rows)
      */
-    selectedRows: PropTypes.arrayOf(PropTypes.object),
+    selectedRows: PropTypes.oneOfType([
+        PropTypes.arrayOf(PropTypes.object),
+        PropTypes.object,
```
alexcjohnson, line 609: Let's spell out all the options here, something like:
```js
    /**
     * The actively selected rows from the grid (may include filtered rows)
     * Can take one of three forms:
     * (1) an array of row objects - if you have defined `getRowId`, you only need the fields it uses.
     * (2) an object containing `function` with a function string (say something about how to construct this)
     * (3) an object containing `ids` with a list of row IDs
     */
    selectedRows: PropTypes.oneOfType([
        PropTypes.arrayOf(PropTypes.object),
        PropTypes.exact({function: PropTypes.string.isRequired}),
        PropTypes.exact({ids: PropTypes.arrayOf(PropTypes.string).isRequired}),
    ]),
```

alexcjohnson:
What happens if you provide a selection as one of the object forms and then the user alters the selection? Does it turn into the array form? If so, does it stay an object until the user makes a change or is it converted to the array form right away?

BSd3v:
Once the selection is processed, it is converted into the response from the grid.

src/lib/fragments/AgGrid.react.js
```diff
@@ -192,15 +194,66 @@ export default class DashAgGrid extends Component {
 
     setSelection(selection) {
         const {gridApi} = this.state;
+        const {getRowId} = this.props;
         if (gridApi && selection) {
-            if (!selection.length) {
+            this.setState({pauseSelections: true});
+            if (Object.keys(selection).includes('function')) {
```
alexcjohnson, line 200:
```suggestion
            if (has('function', selection)) {
```
Just a little cleaner & faster :)

src/lib/fragments/AgGrid.react.js
```diff
@@ -192,15 +194,66 @@ export default class DashAgGrid extends Component {
...
+                    ...customFunctions,
+                    ...window.dashAgGridFunctions,
+                };
+                const test = function (node) {
+                    if (evaluate(parsedCondition, {node, ...context})) {
+                        node.setSelected(true);
+                    }
+                };
+                gridApi.forEachNode(test);
+            } else if (Object.keys(selection).includes('ids')) {
```
alexcjohnson, line 216:
```suggestion
            } else if (has('ids', selection)) {
```

src/lib/fragments/AgGrid.react.js
```diff
@@ -192,15 +194,66 @@ export default class DashAgGrid extends Component {
...
+                    d3,
+                    ...customFunctions,
+                    ...window.dashAgGridFunctions,
+                };
+                const test = function (node) {
+                    if (evaluate(parsedCondition, {node, ...context})) {
+                        node.setSelected(true);
+                    }
+                };
+                gridApi.forEachNode(test);
```
alexcjohnson, line 215: Does it not work to use `this.parseFunction` here?
```suggestion
                const test = this.parseFunction(selection.function);
                gridApi.forEachNode((node) => {
                    if (test(node)) {
                        node.setSelected(true);
                    }
                });
```

alexcjohnson: OIC `parseFunction` puts its argument into `params` and you want it in `node`. That name could be added as an optional argument to `parseFunction` so we can still reuse the rest of that.
```js
parseFunction = memoizeWith(String, (funcString, argName = 'params') => {
...
    return (params) => evaluate(parsedCondition, {[argName]: params, ...context});
```
then you'd call it with:
```js
const test = this.parseFunction(selection.function, 'node');
```

BSd3v: I think we could just leave it as `params`, that way there is unity among all of our functions. XD

alexcjohnson: Haha yeah actually, seems like `params` is more consistent with other functions anyway, so https://github.com/plotly/dash-ag-grid/pull/164/commits/10b361c3b00ab0383b88d565da8e8273f9658207 is good 🎉

alexcjohnson (approved):
💃

PR #174, opened 2023-05-04
floating-columnState

BSd3v:
`columnState` floats during grid interaction and only gets pushed when sent in a callback
`columnDefs` trumps `columnState` if it is pushed in a callback without a `columnState`
`Markdown` renderer now will pass back '' instead of `undefined` if there isnt a value

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to `dash-ag-grid` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source Dash AG Grid repo
 
+## [unreleased]
+
+### Removed
+
+### Added
+
+### Updated
+- [#174](https://github.com/plotly/dash-ag-grid/pull/174)
+  - `columnState` floats during grid interaction and only gets pushed when sent in a callback
+  - `columnDefs` trumps `columnState` if it is pushed in a callback without a `columnState`
+
+### Fixed
+- [#174](https://github.com/plotly/dash-ag-grid/pull/174)
+  - `Markdown` renderer now will pass back `''` instead of `undefined` if there isnt a value
+
 ## [2.0.0] - 2023-05-02
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source D
 
 ### Fixed
 - [#174](https://github.com/plotly/dash-ag-grid/pull/174)
-  - `Markdown` renderer now will pass back `''` instead of `undefined` if there isnt a value
+  - `Markdown` renderer now displays a blank cell rather than writing `undefined` if there is no value. Fixes [#171](https://github.com/plotly/dash-ag-grid/issues/171)
 
 ## [2.0.0] - 2023-05-02
 
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "homepage": "https://github.com/plotly/dash-ag-grid",
     "main": "dash_ag_grid/dash_ag_grid.min.js",
     "scripts": {
-        "prepublishOnly": "rm -rf lib && babel src --out-dir lib --copy-files --config-file ./.babelrc && rm -rf lib/jl/ lib/*.jl",
+        "prepublishOnly": "rimraf -rf lib && babel src --out-dir lib --copy-files --config-file ./.babelrc && rimraf -rf lib/jl/ lib/*.jl",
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "homepage": "https://github.com/plotly/dash-ag-grid",
     "main": "dash_ag_grid/dash_ag_grid.min.js",
     "scripts": {
-        "prepublishOnly": "rm -rf lib && babel src --out-dir lib --copy-files --config-file ./.babelrc && rm -rf lib/jl/ lib/*.jl",
+        "prepublishOnly": "rimraf -rf lib && babel src --out-dir lib --copy-files --config-file ./.babelrc && rimraf -rf lib/jl/ lib/*.jl",
```
alexcjohnson, line 15: Looks like we have `rimraf` installed as a subdependency but that's fragile, let's list it explicitly in `devDependencies`.

```
~/plotly/dash-ag-grid> npm ls rimraf
dash-ag-grid@2.0.0 /Users/alex/plotly/dash-ag-grid
├─┬ eslint@8.39.0
│ └─┬ file-entry-cache@6.0.1
│   └─┬ flat-cache@3.0.4
│     └── rimraf@3.0.2
└─┬ react-docgen@5.4.3
  └─┬ estree-to-babel@3.2.1
    └─┬ c8@7.13.0
      └── rimraf@3.0.2 deduped

```

CHANGELOG.md
```diff
@@ -4,6 +4,21 @@ All notable changes to `dash-ag-grid` will be documented in this file.
...
+### Added
+
+### Updated
+- [#174](https://github.com/plotly/dash-ag-grid/pull/174)
+  - `columnState` floats during grid interaction and only gets pushed when sent in a callback
+  - `columnDefs` trumps `columnState` if it is pushed in a callback without a `columnState`
+
+### Fixed
+- [#174](https://github.com/plotly/dash-ag-grid/pull/174)
+  - `Markdown` renderer now will pass back `''` instead of `undefined` if there isnt a value
```
alexcjohnson, line 20: it passes back `null`, not `''` - but that's an implementation detail, what we want to describe here is the effect of this for users, perhaps:
```suggestion
  - `Markdown` renderer now displays a blank cell rather than writing `undefined` if there is no value. Fixes [#171](https://github.com/plotly/dash-ag-grid/issues/171)
```

tests/test_column_state.py
```diff
@@ -152,4 +231,15 @@ def display_column_state(col_state):
     until(lambda: json.dumps(testState) in dash_duo.find_element('#reset-column-state-grid-pre').text, timeout=3)
 
     dash_duo.find_element('#load-column-state-button').click()
+    until(lambda: json.dumps(colState) in dash_duo.find_element('#reset-column-state-grid-pre').text, timeout=3)
+
+    (
+        ActionChains(dash_duo.driver)
+            .move_to_element(dash_duo.find_element('#load-column-defs'))
+            .click()
+    ).perform()
```
alexcjohnson, line 240: Is there some reason it doesn't work to just click the button? If it's scrolled off-screen maybe we can just put it on a new line or shorten the labels or something?
```suggestion
    dash_duo.find_element('#load-column-defs').click()
```

BSd3v: Clicking it is too fast for some reason. 

There has to be a slight pause between clicks or it doesn't work.

tests/test_markdown_components.py
```diff
@@ -71,6 +71,28 @@ def test_mc001_markdown_components(dash_duo):
...
+            "image": "{0} {0} {0} {0}".format(
+                "![alt text: sun](https://www.ag-grid.com/example-assets/weather/sun.png)"
+            ),
+        },
+        {
+            "make": "***Porsche*** in both",
+            "model": "<b>Boxster</b> in HTML bold",
+            "image": "![alt text: rain](https://www.ag-grid.com/example-assets/weather/rain.png)",
+        },
+    ]
```
alexcjohnson, line 94:
```suggestion
    rowData2 = [{k:row[k] for k in row if k != "link"} for row in rowData]
```

tests/test_column_state.py
alexcjohnson: ok - maybe just add a little `time.sleep` then, with a comment about why?

BSd3v: I was trying to avoid that. Haha. 

But yes, a time.sleep works.

alexcjohnson: haha yeah `time.sleep` is generally not great, but if there's a real reason we need a delay rather than waiting for a specific event, it's better to do that explicitly rather than something implicit like this.

alexcjohnson (approved):
💃 Looks great, just the last minor comment about `time.sleep` then we can merge. Very nice fixes and tests!

PR #186, opened 2023-05-11
Docs review - batch 2

sdidier-dev:
Hello !
Next batch of docs review 📝
Today, pages:
- Column sizing
- Column moving
- Column pinning

\+ Minor modifications of pages Column headers and Column state

### DIFF
--- a/docs/assets/columns.css
+++ b/docs/assets/columns.css
@@ -1,79 +1,104 @@
 .header1 .ag-header-cell-label {
-  /*Necessary to allow for text to grow vertically*/
-  height: 100%;
-  padding: 0 !important;
+    /*Necessary to allow for text to grow vertically*/
+    height: 100%;
+    padding: 0 !important;
 }
 
 .header1 .ag-header-group-cell {
-  font-size: 50px;
+    font-size: 50px;
 }
 
 .header1 .ag-header-cell-label .ag-header-cell-text {
-  /*Force the width corresponding at how much width
-    we need once the text is laid out vertically*/
-  width: 55px;
-  writing-mode: vertical-lr;
-  -ms-writing-mode: tb-lr;
-  line-height: 2em;
-  margin-top: 60px;
+    /*Force the width corresponding at how much width
+      we need once the text is laid out vertically*/
+    width: 55px;
+    writing-mode: vertical-lr;
+    -ms-writing-mode: tb-lr;
+    line-height: 2em;
+    margin-top: 60px;
 }
 
 /*Floating filters*/
 .header1 .ag-floating-filter-body {
-  height: 50px;
+    height: 50px;
 }
 
 .header1 .ag-floating-filter-body input {
-  height: 49px;
+    height: 49px;
 }
 
 .header1 .ag-floating-filter-button {
-  margin-top: -49px;
+    margin-top: -49px;
 }
 
 .header1 .ag-floating-filter-button button {
-  height: 49px;
+    height: 49px;
 }
 
 .header1 .ag-floating-filter-body input {
-  font-size: 15px;
-  font-weight: bold;
+    font-size: 15px;
+    font-weight: bold;
 }
 
 .header3 .ag-header-group-cell-with-group {
-  background-color: #00e7b1 !important;
+    background-color: #00e7b1 !important;
 }
 
 /* Used in Column Groups page for Group Changes example*/
 .participant-group {
-  background-color: #00e7b1 !important;
+    background-color: #00e7b1 !important;
 }
+
 .medals-group {
-  background-color: #5b9bd5 !important;
+    background-color: #5b9bd5 !important;
+}
+
+/* Used in Column Sizing page for Resizing Groups example*/
+.fixed-size-header {
+    background-color: #ffdddd !important;
+}
+
+.resizable-header {
+    background-color: #ddffdd !important;
+}
+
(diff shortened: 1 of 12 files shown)


BSd3v:
> For the first example, I tried to use something like defaultState as in [AG Grid example](https://www.ag-grid.com/react-data-grid/column-pinning/#example-pinning), instead of updating the pinned state of all cols, but couldn't find anything.
Moreover, while investigating, I found in AgGrid component the parameter updateColumnState, that triggers AG Grid
method updateColumnState() when set to True, but I didn't find this method in [AG Grid docs](https://www.ag-grid.com/react-data-grid/column-api/#reference-state)

To set a default state, you would pass it inside the `columnState` prop, as when passed it performs the `applyColumnState`, so anything passed to it will be applied like in the docs:

`columnState={ defaultState: { pinned: None} }`

As far as `updateColumnState`, we removed it from the docs as this is a catch all for snagging the `columnState` from the grid, this may be obsolete due to changes we made when we released 2.0.0, but I insisted on keeping it incase the `columnState` =/= the grid's current columnState.

AnnMarieW:
@sdidier-dev Thanks for another great pull request :partying_face: 

A couple comments:

__Column Headers__
With the `floatingFilterHeight` set to  25, you can't see the filter icon  next to the filter input box -- it's still there, it's just not visible.  It makes it unclear that it's a floating filter row. 

image

Here's a floating filter with the icon
image

I was unable to align the icon.  If you know a way, that would be awesome -- otherwise we should remove the floating filter for now.

__Column Sizing__
I like your changes in this section.  I'm certainly open to a major overhaul.  The way Dash handles the column sizes is different than the upstream AG Grid component, so anything to make this section more understandable will be great.  

For the code formatting, we use [Black](https://black.readthedocs.io/en/stable/).  Even though I don't always agree with how it formats, it does make the code very consistent.  No need to change anything in your pull request. Just know that after it's merged, some formatting will change when we run Black. 

All the other changes to other sections look fantastic.  

Just let me know about the floating filter height.  One option is to take that out for now so it doesn't hold up all these other great changes.  :tada:

sdidier-dev:
Hi @AnnMarieW and @BSd3v !

@BSd3v,  
Thank you for this explanation ! Full of hope, I ran some tests.  
As you suggested, I tried to provide the argument `columnState={'defaultState': {'pinned': None}}`, but it expects an
array:

```
Invalid argument `columnState` passed into AgGrid with ID "column-pinning".
Expected `array`.
Was supplied type `object`.
```

What I can understand, the applyColumnState should have the form:

```
.applyColumnState({
    state: [{ colId: 'country', pinned: 'left' }],
    defaultState: { pinned: null },
})
```

and the dash AG Grid component seems to provide the value only for the `state` key, meaning something like
`columnState=[{ 'colId': 'country', 'pinned': 'left' }]` will perform

```
.applyColumnState({
    state: [{ colId: 'country', pinned: 'left' }],
})
```

So I still can't find a way to use `defaultState`. Do I miss something?  
It is not really important right now, I think the example works fine as is.

@AnnMarieW,  
Weird, on my side, the icon is still there, even with a smaller height (I tried 15px):  
image

Whatever, I wanted to *"provide a unique look and feel"*, but who will do this? 😅  
I'm going to set the floating filter height to 50px like in [AG Grid docs](https://www.ag-grid.com/react-data-grid/column-headers/#example-header-height-and-text-orientation)
I think it is still useful to keep it in this example, because if we use a floating filter and set the `headerHeight`
and not `floatingFiltersHeight`, it seems that the later will be equal to `headerHeight`. Here, meaning 150px...

I don't know Black for the code formatting, but I'm going to take a look.
And I will of course check the result of the formatting once merged.

About the community forum, I'm definitely going to use it, even if I need to use Edge as long as I didn't solve my issue 😀

sdidier-dev:
Here we go!

I agree with you @BSd3v (for what I have seen until now)👍 
It was mainly to stick with the AG Grid docs.

AnnMarieW:
@sdidier-dev 
Strange the filter icon works for you. I'll merge as-is and we'll see how it looks live.

Just a heads up  - the Plotly folks are making great progress moving these docs over to the official Dash Docs.  If you are willing to continue your review, we'd still appreciate your input.   All the content and example apps are being moved over.  There will likely be more editing on the content and formatting like headings etc once it's on the Plotly site :-)

AnnMarieW:
The changes are live - but they still look odd.  How do they look for you?  https://dashaggrid.pythonanywhere.com/columns/column-headers

image

sdidier-dev:
Yes, I saw that too! I think I know the issue, it seems to be the .css formatting.
I 'm going to take a look

sdidier-dev:
Found it! 
In columns.css:
```css
/*Floating filters*/
.header1 .ag-floating-filter-body {
    height: 50px;
}

.header1 .ag-floating-filter-body input {
    height: 49px;
}

.header1 .ag-floating-filter-button {
    margin-top: -49px;
}

.header1 .ag-floating-filter-button button {
    height: 49px;
}

.header1 .ag-floating-filter-body input {
    font-size: 15px;
    font-weight: bold;
}
```

It probably comes from [AG Grid docs](https://www.ag-grid.com/react-data-grid/column-headers/#example-text-orientation) in their style.css (they don't even use floating filters in the example...)

I checked, it seems that the class `.header1` is not used elsewhere in dash_ag_grid module, so it should be safe to remove them  
I'm going to make the modification

PR #188, opened 2023-05-11
Fix example in Column Headers: Remove .css formatting of floating filters

sdidier-dev:
Should be better now !


AnnMarieW:
Yes, looks great now, thanks! :woman_dancing:

Issue #179, opened 2023-05-08
[Feature Request] - scroll to row


AnnMarieW:
Scrolling to a row is a frequently requested feature for DataTable.  It looks like it available in AG Grid: https://www.ag-grid.com/javascript-data-grid/grid-api/#reference-scrolling

PR #199, opened 2023-05-26
Add prop `scrollToRow`

Farkites:
Adds the prop `scrollTo` which takes a `dict` with the following keys:
- `rowIndex`: row number to scroll to.
- `rowId`: row id to scroll to.
- `rowPosition`: position of the row in the grid after scrolling. Can be `top`, `middle`, or `bottom`.
- `column`: column to scroll to.
- `columnPosition`: position of the column in the grid after scrolling. Can be `auto`, `start`, `middle`, or `end`.

#179


BSd3v:
Hello @Farkites,

I think this is a start, I think instead of limiting to just `scrollToRow`, it should be `scrollTo` where you can pass different keys to perform different actions, like you want to scroll to a specific cell and make it active, then you pass coordinates, etc.

Farkites:
Hello @BSd3v , 
Thanks for the suggestion. I've modified the prop to `scrollTo` and added the option to scroll to columns and the desired position of the rows and columns after scrolling. I appreciate any feedback you may have about this

BSd3v:
Looks much better, however you are still missing the `node` version of the finding the row.

https://www.ag-grid.com/react-data-grid/grid-api/#reference-scrolling-ensureNodeVisible

Also, please add a test for this using all three, and combinations of them as well. 🙂 

Probably just 5 uses of callbacks.

BSd3v:
Adding a `rowId` works when there is an `getRowId` provided, but when there is none provided, and the dev doesnt want to take a guess at the `rowIndex`, we need to be able to pass the `node.data` that needs to be scrolled to.

src/lib/components/AgGrid.react.js
BSd3v, line 1: need to add `node` or `data` as a prop here, for use-cases when the dev doesnt have either `rowIndex` or `rowId`

src/lib/fragments/AgGrid.react.js
```diff
@@ -1010,6 +1017,38 @@ export default class DashAgGrid extends Component {
...
+                scrollTo.rowPosition === null ? 'top' : scrollTo.rowPosition;
+            gridApi.ensureIndexVisible(scrollTo.rowIndex, rowPosition);
+        }
+        if (scrollTo.rowId) {
+            const rowPosition =
+                scrollTo.rowPosition === null ? 'top' : scrollTo.rowPosition;
+
+            const node = gridApi.getRowNode(scrollTo.rowId);
+            gridApi.ensureNodeVisible(node, rowPosition);
+        }
```
BSd3v, line 1037: add `node` or `data` here, also, I'd recommend an else if instead of ifs.

For example, say a dev passed three different rows for the grid to scroll to, `rowIndex`, `rowId` and `data`, this would the finally end up on data, but could cause some interesting glitches.

something like: `if rowIndex else if rowId else if node/data`

Bring the `rowPosition` declaration outside of the ifs, since this isnt going to change.

I dont believe that you need the `=== null`, this seems redundant.

tests/test_scroll_to.py
BSd3v, line 1: This test is definitely a start, but you arent really testing that the grid is scrolling to those points, or rendering in the certain fashion. You are just testing that your prop is adjusting properly.

BSd3v (changes requested):
@alexcjohnson/ @AnnMarieW , what are your thoughts on the naming convention of the `rowIndex`, `rowId`, `node/data`?

src/lib/fragments/AgGrid.react.js
```diff
@@ -1023,23 +1023,23 @@ export default class DashAgGrid extends Component {
...
-            const rowPosition =
-                scrollTo.rowPosition === null ? 'top' : scrollTo.rowPosition;
-
+        } else if (scrollTo.rowId) {
             const node = gridApi.getRowNode(scrollTo.rowId);
             gridApi.ensureNodeVisible(node, rowPosition);
+        } else if (scrollTo.data) {
+            gridApi.forEachNode((node) => {
+                if (includes(node.data, [scrollTo.data])) {
+                    gridApi.ensureNodeVisible(node, rowPosition);
```
BSd3v, line 1035: find some way to break it once it finds the first match, this way you dont have to keep going through the whole dataset

BSd3v (commented):
Looks good, make sure your test also includes providing an initial scrollTo to see if that it working as well.

BSd3v (commented):
Change the first test to use the same browser session and then make sure that cells around the scrolled to cell are not visible based upon the position arguments.

Also, make sure to merge dev into the branch and update the changelog entries.

tests/test_scroll_to.py
```diff
@@ -0,0 +1,147 @@
...
+    dash_duo.find_element("#btn").click()
+    dash_duo.wait_for_text_to_equal(
+        "#scrollTo-output", json.dumps(scroll_to_input), timeout=5
+    )
+
+    until(
+        lambda: x == grid.get_cell(row, column).location["x"]
+        and y == grid.get_cell(row, column).location["y"],
+        timeout=3,
+    )
```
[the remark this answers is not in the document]
Farkites: Thanks for the suggestion! But `isDisplayed()` is returning `True` for rows that are not actually displayed but 10 rows below the last one supposed to be displayed.

src/lib/fragments/AgGrid.react.js
```diff
@@ -1042,6 +1049,38 @@ export default class DashAgGrid extends Component {
...
+        const {scrollTo, setProps} = this.props;
+        if (!gridApi) {
+            return;
+        }
+        const rowPosition = scrollTo.rowPosition ? scrollTo.rowPosition : 'top';
+        if (scrollTo.rowIndex || scrollTo.rowIndex === 0) {
+            gridApi.ensureIndexVisible(scrollTo.rowIndex, rowPosition);
+        } else if (scrollTo.rowId) {
+            const node = gridApi.getRowNode(scrollTo.rowId);
+            gridApi.ensureNodeVisible(node, rowPosition);
```
alexcjohnson, line 1063: I'd probably have `rowId` take precedence over `rowIndex`, though I don't know why you'd provide both. Maybe you want to get a certain row but it might not exist you want an index to fall back on?

also `rowId` could be falsy, so `if (scrollTo.rowId !== undefined)` perhaps?

src/lib/fragments/AgGrid.react.js
```diff
@@ -1042,6 +1049,38 @@ export default class DashAgGrid extends Component {
...
+        } else if (scrollTo.data) {
+            gridApi.forEachNode((node) => {
+                if (includes(node.data, [scrollTo.data])) {
```
alexcjohnson, line 1066:
```suggestion
                if (equals(node.data, scrollTo.data)) {
```
But if there's a `getRowId` we can do better, right?
```
const node = gridApi.getRowNode(getRowId(scrollTo.data));
gridApi.ensureNodeVisible(node, rowPosition);
```

src/lib/fragments/AgGrid.react.js
alexcjohnson: @BSd3v you resolved this but I think we can do better, even if `forEachNode` doesn't have a way to short-circuit:
(1) outside the loop `let scrolled = false;`, then inside the loop `if (!scrolled && equals(node.data, scrollTo.data)) { scrolled = true; ... }`
(2) instead of `forEachNode` we should use I think `forEachNodeAfterFilterAndSort` because there's no point scrolling to a hidden node.

src/lib/fragments/AgGrid.react.js
Farkites: My thought process on the priority of `rowIndex`, `rowId`, and `data` was to go from more to less specific (`rowIndex` always have unique values, but `rowId` can be set to have non-unique values). I think that the use case for using both deliberately is pretty niche, but with the current implementation it wouldn't fall back on any index/rowId because of the `if-else`.

BSd3v: What happens if the grid can't scroll to a row? I think it returns false. 

Could use the scrolled idea from your other suggestion @alexcjohnson.

So, instead of doing if-else. They are all if with the additional condition of if it scrolls. 

rowIds being non-unique is an issue that causes a great deal of pain, even outside of scrolling. Sorting creates multiple lines, etc. We should expect the rowId to be unique.

src/lib/fragments/AgGrid.react.js
Farkites: To use `getRowId` we would have to always specify the function in the prop. It would be very similar to just using `rowId`

BSd3v: getRowId exists, and we use it in the setSelection for the selectedRows. This is without defining it.

BSd3v: Also, equals will not work in this instance. I believe that node.data has a few extra keys compared to the regular data from the df.

alexcjohnson: @Farkites has already taken the suggestion, but @BSd3v `equals` is what was happening here anyway with `includes`. From https://ramdajs.com/docs/#includes:
> Returns `true` if the specified value is equal, in [`R.equals`](https://ramdajs.com/docs/#equals) terms, to at least one element of the given list; `false` otherwise.

So my suggestion was an exact replacement (and apparently works, or the tests would have failed!)

alexcjohnson: We can leave it as is for now, but I can still see a purpose using `data` when there's a `getRowId` defined, and accelerating the lookup with the ID. Maybe you know which fields are involved in the ID but the function to put them together is awkward - especially since you're likely constructing it in Python but the function is normally executed in JS where there could be subtle differences.

src/lib/fragments/AgGrid.react.js
alexcjohnson: Yeah OK - I was thinking about things like you've filtered out the row you're trying to scroll to, maybe instead you want to scroll to the top - so that would be ID (or data) falling back on index. But reflecting on this, most likely in that case I'd rather the grid not scroll at all, in which case this doesn't matter.

alexcjohnson (approved):
💃

src/lib/fragments/AgGrid.react.js
BSd3v: I think we can still use the getRowId, beats going through an entire set of data if we can just jump to the node.

BSd3v (approved):
💃

PR #200, opened 2023-05-26
Column visibility

Farkites:
- Adds `columnVisible` prop and `onColumnVisible` event.

When using the enterprise tool panel to show/hide a column the `columnVisible` prop is updated with a list of dictionaries with all the columns and their visibility state.


BSd3v:
Hello @Farkites,

What is your thinking on this, why cant you use the `columnState` for this? This information should be listed there.

Farkites:
Hello @BSd3v, I guess that the only advantage of this approach is that this `columnVisible` property is only modified when hiding/showing columns whereas `columnState` gets triggered with many different events.

This can anyways be handled in the callback in all the use cases I can think of right now... Additionally, as shown in the test, my approach times out in some tests, so this approach is just slowing down the component.

Thanks for the comment!

PR #208, opened 2023-06-02
adding support for `cellRenderer` functions

BSd3v:


BSd3v:
No consistent benefit that I can see, may revisit later.

PR #209, opened 2023-06-12
exposing the api and columnApi to clientside js

BSd3v:
adding functions `getApi` and `getColumnApi` to the `dash_ag_grid` namespace

### DIFF
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -1,4 +1,5 @@
 /* eslint-disable import/prefer-default-export */
 import AgGrid from './components/AgGrid.react';
+import {getApi, getColumnApi} from './fragments/AgGrid.react';
 
-export {AgGrid};
+export {AgGrid, getApi, getColumnApi};
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -1294,6 +1294,26 @@ export default class DashAgGrid extends Component {
     }
 }
 
+export const getApi = (id) => {
+    var grid = agGridRefs[stringifyId(id)];
+    if (grid?.api) {
+        return grid.api;
+    }
+    throw new Error(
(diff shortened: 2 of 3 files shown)


src/lib/index.js
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable import/prefer-default-export */
 import AgGrid from './components/AgGrid.react';
+import {getApi, getColumnApi} from './fragments/AgGrid.react';
```
alexcjohnson, line 3: Hmm this is awkward... we can't just import from the fragment or it ceases to be an async chunk.

Simplest thing that comes to mind: put the outer functions in `components/AgGrid.react` along with a container you can inject inner functions into:
```js
export const apiGetters = {};

const _get = (flavor) => (id) => {
    // optional chaining so before the fragment exists it'll just return undefined
    // which does the right thing because clearly no grid is initialized yet!
    const api = apiGetters[flavor]?.(id);
    if (api) {
        return api;
    }
    throw new Error(
        `no grid found, or grid is not initialized yet, with id: ${id}`
    );
};
export const getApi = _get('getApi');
export const getColumnApi = _get('getColumnApi');
```
Then in the fragment, inject the real API getters:
```js
import {apiGetters} from '../components/AgGrid.react'; // combine with propTypes and defaultProps import we already have

apiGetters.getApi = (id) => agGridRefs[stringifyId(id)]?.api;
apiGetters.getColumnApi = (id) => agGridRefs[stringifyId(id)]?.columnApi;
```

src/lib/index.js
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable import/prefer-default-export */
```
alexcjohnson, line 1: There are multiple exports now, so we can probably delete this?

alexcjohnson (approved):
💃 looks good! Just a changelog entry and we can merge 🎉

PR #210, opened 2023-06-15
fixing `columnState` for use with React 18

BSd3v:
* migrated props that use `setProps` from `render()` to `componentDidUpdate` (this fixed the `bad setState()` error)
* adjusted `columnState` if statement to use `prevProps.loading_state`


alexcjohnson:
This looks great. I wonder though, do we need to call out to these same updaters in `componentDidMount`? Like what if you want your grid to start out scrolled to a specific row?

BSd3v:
The grid can't do things like that until it is ready, which updates the grid's state. Which then triggers the event if the prevState had a null gridApi and the current one has a gridApi.

BSd3v:
Also, there is already a test for starting out on a scrolled row, as well as with an initial columnState.

alexcjohnson (approved):
💃

PR #212, opened 2023-06-21
allow `dash_ag_grid` api to be an async function

BSd3v:
* allows for rendering to take place and to wait until it is available

### DIFF
--- a/src/lib/components/AgGrid.react.js
+++ b/src/lib/components/AgGrid.react.js
@@ -730,6 +730,9 @@ export const defaultProps = DashAgGrid.defaultProps;
 
 export const apiGetters = {};
 
+const DEFAULTTRYCOUNT = 20;
+const DEFAULTPAUSE = 200;
+
 const _get = (flavor) => (id) => {
     // optional chaining so before the fragment exists it'll just return undefined
     // which does the right thing because clearly no grid is initialized yet!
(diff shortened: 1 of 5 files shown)


src/lib/components/AgGrid.react.js
```diff
@@ -741,5 +744,30 @@ const _get = (flavor) => (id) => {
...
+const _getAsync =
+    (flavor) =>
+    async (id, trycount = DEFAULTTRYCOUNT) => {
+        // optional chaining so before the fragment exists it'll just return undefined
+        // which does the right thing because clearly no grid is initialized yet!
+        var api = apiGetters[flavor]?.(id);
+        const delay = (ms) => new Promise((res) => setTimeout(res, ms));
+        let count = 0;
+        while (!api) {
+            await delay(DEFAULTPAUSE);
```
alexcjohnson, line 756: Might be better to implement exponential backoff... something like start at 1ms and increase 50% each step:
```suggestion
        let pause = 1;
        while (!api) {
            await delay(pause);
            pause *= 1.5;
```
If you do it that way, with 20 steps the last step is ~3sec (total time ~9sec) but with a factor of 1.5 you're guaranteed to never wait more than 50% longer than it took for the API to resolve, no matter how long that was, so if it resolves quickly it's snappier than always needing to wait 200ms.

alexcjohnson (approved):
💃 after considering my comment about exponential backoff. Great feature!

PR #225, opened 2023-07-25
bumping AG Grid for v30

BSd3v:
- fixing breaking changes in coding


src/lib/fragments/AgGrid.react.js
```diff
@@ -385,6 +382,15 @@ export default class DashAgGrid extends Component {
                     return c;
                 });
             }
+            if (OBJ_MAP_MAYBE_FUNCTIONS[target]) {
+                Object.keys(value).map((c) => {
+                    console.log(c);
```
AnnMarieW, line 387: remove log :-)

AnnMarieW:
A couple questions/requests

- Is it possible to take advantage of the new feature [providing modules to individual grids](https://www.ag-grid.com/archive/30.0.0/react-data-grid/modules/?ref=blog.ag-grid.com#providing-modules-to-individual-grids) ?
This would be helpful in the docs to mix enterprise and community examples on the same page.

- Allow the following props to accept functions:
   -  `dataTypeMatcher`
    -  `dateParser`
   - `dateFormatter`

These are used in the [dataTypeDefinitions.](https://www.ag-grid.com/react-data-grid/grid-options/#reference-columns-dataTypeDefinitions)  I tried using the API, but there seems to be timing differences as this needs to be applied when the grid is first rendered.  For examples of the use-case see [Overridding the pre-defined cell data type definitions](https://www.ag-grid.com/react-data-grid/cell-data-types/#overriding-the-pre-defined-cell-data-type-definitions)

BSd3v:
Have you tried both of them together?

We can try... Try adding those things to the propCategories.

PR #226, opened 2023-07-25
fixing row-grouping editing

BSd3v:
- also fixed `virtualRowData` to only include nodes where there is data

### DIFF
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -471,16 +471,9 @@ export default class DashAgGrid extends Component {
 
     getRowData() {
         const newRowData = [];
-        const {openGroups} = this.state;
-        if (!isEmpty(openGroups)) {
-            this.state.gridApi.forEachLeafNode((node) => {
-                newRowData.push(node.data);
-            });
-        } else {
-            this.state.gridApi.forEachNode((node) => {
-                newRowData.push(node.data);
-            });
-        }
+        this.state.gridApi.forEachLeafNode((node) => {
+            newRowData.push(node.data);
+        });
         return newRowData;
     }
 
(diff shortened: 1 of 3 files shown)


src/lib/fragments/AgGrid.react.js
```diff
@@ -459,7 +459,9 @@ export default class DashAgGrid extends Component {
         if (rowModelType === 'clientSide') {
             const virtualRowData = [];
             this.state.gridApi.forEachNodeAfterFilterAndSort((node) => {
-                virtualRowData.push(node.data);
+                if (node.data) {
```
alexcjohnson, line 462: what's the condition where you get falsy `node.data`?

BSd3v: If the rows are grouped, they put undefined because the row is only for displaying the data in the grid. 

You can see it if you group rows and print out the virtualRowData.

src/lib/fragments/AgGrid.react.js
```diff
@@ -469,9 +471,16 @@ export default class DashAgGrid extends Component {
...
-        });
+        const {openGroups} = this.state;
+        if (!isEmpty(openGroups)) {
+            this.state.gridApi.forEachLeafNode((node) => {
+                newRowData.push(node.data);
+            });
+        } else {
+            this.state.gridApi.forEachNode((node) => {
+                newRowData.push(node.data);
+            });
```
alexcjohnson, line 482: Why is `!isEmpty(openGroups)` the condition to use `forEachLeafNode`? Would it work to just always use `forEachLeafNode` in this function?

src/lib/fragments/AgGrid.react.js
alexcjohnson: But that doesn't apply to `forEachNode` or `forEachLeafNode`?

BSd3v: `forEachNode` it is an issue. Not with `forEachLeafNode`, but neither of these options take into account the sort or filter of the grid. So it has to be tested for.

src/lib/fragments/AgGrid.react.js
BSd3v: I changed it, all the tests passed, so I guess it is capable. 🤞

alexcjohnson (approved):
💃

Issue #237, opened 2023-08-31
[Bug] Components not working in the Detail Grid


rpforrest1:
I am wanting to use cell renderers for the Detail Grid. In the example below, I am unable to get markdown to render inside a Detail Grid. I hope to expand this out to use any cellRenderer (not just markdown).

Below is an example to reproduce the issue.

```
from dash import Dash, html
import dash_ag_grid as dag

import os

app = Dash(__name__)

masterColumnDefs = [
    {
        "headerName": "Country",
        "field": "country",
        "cellRenderer": "agGroupCellRenderer",
        "cellRendererParams": {
            'innerRenderer': "markdown",
        },
    },
    {
        "headerName": "Region", 
        "field": "region"
    },
    {
        "headerName": "Population",
        "field": "population"
    },
    {
        "headerName": "Test Link",
        "field": "test_link",
        "cellRenderer": "markdown",
    },
]
detailColumnDefs = [
    {
        "headerName": "City", 
        "field": "city",
        "resizable": True,
        "cellRenderer": "markdown"
    },
    {
        "headerName": "Pop. (City proper)", 
        "field": "population_city",
        "resizable": True,
    },
    {
        "headerName": "Pop. (Metro area)",
        "field": "population_metro",
        "resizable": True,
    },
    {
        "headerName": "Test Render",
        "field": "test_render",
        "resizable": True,
        "cellRenderer": "agGroupCellRenderer",
        "cellRendererParams": {
            'checkbox': True,
            'innerRenderer': "markdown",
        },
    }
]
rowData = [
    {
        "country": "**China**",
        "region": "Asia",
        "population": 1411778724,
        "test_link":"[Blank Test Link](/)",
        "cities": [
            {
                "city": "**Shanghai**",
                "population_city": 24870895,
                "population_metro": "NA",
                "test_render": "**BOLD TEST**"
            },
            {
                "city": "**Beijing**",
                "population_city": 21893095,
                "population_metro": "NA",
                "test_render": "**BOLD TEST**"
            },
            {
                "city": "**Chongqing**",
                "population_city": 32054159,
                "population_metro": "NA",
                "test_render": "**BOLD TEST**"
            },
        ],
    },
    {
        "country": "**India**",
        "region": "Asia",
        "population": 1383524897,
        "test_link":"[Blank Test Link](/)",
        "cities": [
            {
                "city": "**Delhi**",
                "population_city": 16753235,
                "population_metro": 29000000,
[...]

PR #238, opened 2023-08-31
Fixing issue #237

BSd3v:
where grid components werent passed down to detail grids

### DIFF
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -13,6 +13,7 @@ import {
     omit,
     includes,
     assoc,
+    assocPath,
 } from 'ramda';
 import {
     propTypes as _propTypes,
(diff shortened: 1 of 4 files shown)


src/lib/fragments/AgGrid.react.js
```diff
@@ -413,6 +413,9 @@ export default class DashAgGrid extends Component {
                         ? this.suppressGetDetail(value.detailColName)
                         : this.callbackGetDetail;
                 }
+                if ('detailGridOptions' in value) {
+                    value.detailGridOptions.components = this.state.components;
```
alexcjohnson, line 417: We shouldn't be mutating `value` (here or in the block above `value.getDetailRowData = ...`, sorry I hadn't noticed this before) - so we should do something like:
```js
let adjustedVal = value;
if ('suppressCallback' in value) {
    adjustedVal = {
        ...adjustedval,
        getDetailRowData: value.suppressCallback
                        ? this.suppressGetDetail(value.detailColName)
                        : this.callbackGetDetail
    };
}
if ('detailGridOptions' in value) {
    adjustedVal = {
        ...adjustedVal,
        detailGridOptions: {
            ...value.detailGridOptions,
            components: this.state.components
        }
    };
}
```

alexcjohnson: Can also do this with https://ramdajs.com/docs/#assoc and `assocPath` if you prefer, might be more compact.

alexcjohnson (approved):
💃

PR #243, opened 2023-10-12
allowing more functions

BSd3v:
adding filter to be allowed as a functional component without params

### DIFF
--- a/tests/assets/dashAgGridFunctions.js
+++ b/tests/assets/dashAgGridFunctions.js
@@ -225,3 +225,116 @@ dagfuncs.YearFilter = forwardRef((props, ref) => {
         }
         )
 });
+
+dagfuncs.setBody = () => {
+    return document.querySelector('body')
+}
+
+// cell editor custom component  - dmc.Select
+dagfuncs.DMC_Select = class {
+    // gets called once before the renderer is used
+    init(params) {
+        // create the cell
+        this.params = params;
+
+        // function for when Dash is trying to send props back to the component / server
+        var setProps = (props) => {
+            if (typeof props.value != typeof undefined) {
+                // updates the value of the editor
+                this.value = props.value;
+
+                // re-enables keyboard event
+                delete params.colDef.suppressKeyboardEvent;
+
+                // tells the grid to stop editing the cell
+                params.api.stopEditing();
+
+                // sets focus back to the grid's previously active cell
+                this.prevFocus.focus();
+            }
+        };
+        this.eInput = document.createElement('div');
+
+        // renders component into the editor element
+        ReactDOM.render(
+            React.createElement(window.dash_mantine_components.Select, {
+                data: params.options,
+                value: params.value,
+                setProps,
+                style: {width: params.column.actualWidth-2,  ...params.style},
+                className: params.className,
+                clearable: params.clearable,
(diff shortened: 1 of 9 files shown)


tests/assets/dashAgGridFunctions.js
```diff
@@ -171,4 +171,52 @@ function monthToComparableNumber(date) {
   const monthNumber = parseInt(date.split('/')[1]);
   const dayNumber = parseInt(date.split('/')[0]);
   return yearNumber * 10000 + monthNumber * 100 + dayNumber;
-}
\ No newline at end of file
+}
+
+const [useImperativeHandle, useState, useEffect, forwardRef] = [React.useImperativeHandle, React.useState, React.useEffect, React.forwardRef]
```
alexcjohnson, line 176:
```suggestion
const {useImperativeHandle, useState, useEffect, forwardRef} = React;
```

tests/assets/dashAgGridFunctions.js
```diff
@@ -171,4 +171,52 @@ function monthToComparableNumber(date) {
   const monthNumber = parseInt(date.split('/')[1]);
   const dayNumber = parseInt(date.split('/')[0]);
   return yearNumber * 10000 + monthNumber * 100 + dayNumber;
-}
\ No newline at end of file
+}
+
+const {useImperativeHandle, useState, useEffect, forwardRef} = React;
+
+dagfuncs.YearFilter = forwardRef((props, ref) => {
```
alexcjohnson, line 178:
```suggestion
// This example was adapted from https://www.ag-grid.com/react-data-grid/component-filter/
// The only differences are:
// - React.createElement instead of JSX
// - setProps, which all Dash components use to report user interactions,
//     instead of a plain js event handler
dagfuncs.YearFilter = forwardRef((props, ref) => {
```

tests/assets/dashAgGridFunctions.js
```diff
@@ -171,4 +171,52 @@ function monthToComparableNumber(date) {
...
+       }
+   });
+
+   useEffect(() => {
+       props.filterChangedCallback()
+   }, [year]);
+
+    setProps = (props) => {
+        if (props.value) {
+            setYear(props.value)
```
alexcjohnson, line 212:
```suggestion
    setProps = ({value}) => {
        if (value) {
            setYear(value)
```

tests/test_popupparent.py
```diff
@@ -0,0 +1,36 @@
...
+
+    grid = utils.Grid(dash_duo, "grid")
+
+    grid.wait_for_cell_text(0, 0, "")
+
+    ### testing animations
+    action = utils.ActionChains(dash_duo.driver)
+    action.double_click(grid.get_cell(0, 0)).perform()
+
+    dash_duo.find_element('body > .ag-popup')
```
alexcjohnson, line 36: Nice test - and nice use of `>` in the selector to verify that it's a direct child! Do you want to include something to assert that there's a dmc.Select in the popup? Like `dash_duo.find_element('body > .ag-popup .mantine-select')` or whatever class it should have?

BSd3v: Yeah, we can definitely do this.

alexcjohnson (approved):
💃 looks great!

PR #246, opened 2023-10-16
Bump to v31

BSd3v:
Reopening to go onto `main` since `dev` was deleted


AnnMarieW:
I'm trying to get [This example](https://www.ag-grid.com/react-data-grid/cell-data-types/#overriding-the-pre-defined-cell-data-type-definitions) from the AG Grid docs working, but I'm not sure if I'm defining it wrong or if `dataTypeDefinitions` prop needs to have function enabled.


```

import dash_ag_grid as dag
from dash import Dash, html
import pandas as pd


app = Dash(__name__)

df = pd.read_csv(
    "https://raw.githubusercontent.com/plotly/datasets/master/ag-grid/olympic-winners.csv"
)

columnDefs = [
    {"field": 'athlete'},
    {"field": 'age'},
    {"field": 'date'},
]
defaultColDef = {
      "filter": True,
      "floatingFilter": True,
      "editable": True,
}

app.layout = html.Div(
    [
        dag.AgGrid(
            id="grid-cell-data-types",
            columnDefs=columnDefs,
            defaultColDef=defaultColDef,
            dashGridOptions={"dataTypeDefinitions": {"function": "dataTypeDefinitions"}},
        ),
    ],
)

app.run(debug=True)



"""

dagfuncs.dataTypeDefinitions = {
      dateString: {
        baseDataType: 'dateString',
        extendsDataType: 'dateString',
        valueParser: (params) =>
          params.newValue != null &&
          params.newValue.match('\\d{2}/\\d{2}/\\d{4}')
            ? params.newValue
            : null,
        valueFormatter: (params) => (params.value == null ? '' : params.value),
        dataTypeMatcher: (value) =>
          typeof value === 'string' && !!value.match('\\d{2}/\\d{2}/\\d{4}'),
        dateParser: (value) => {
          if (value == null || value === '') {
            return undefined;
          }
          const dateParts = value.split('/');
          return dateParts.length === 3
            ? new Date(
                parseInt(dateParts[2]),
                parseInt(dateParts[1]) - 1,
                parseInt(dateParts[0])
              )
            : undefined;
        },
        dateFormatter: (value) => {
          if (value == null) {
            return undefined;
          }
          const date = String(value.getDate());
          const month = String(value.getMonth() + 1);
          return `${date.length === 1 ? '0' + date : date}/${
            month.length === 1 ? '0' + month : month
          }/${value.getFullYear()}`;
        },
      },
    };


"""


```

AnnMarieW:
Is the `date` object [cell data type](https://www.ag-grid.com/javascript-data-grid/cell-data-types/) applicable in Dash?  If so, how would you use it?

AnnMarieW:
There are a couple warning messages in the console in V31

AG Grid: Since v31, 'columnApi.getColumnState' is deprecated and moved to 'api.getColumnState'.

AG Grid: Since v31, 'columnApi.autoSizeAllColumns' is deprecated and moved to 'api.autoSizeAllColumns'.

BSd3v:
> There are a couple warning messages in the console in V31
> 
> AG Grid: Since v31, 'columnApi.getColumnState' is deprecated and moved to 'api.getColumnState'.
> 
> AG Grid: Since v31, 'columnApi.autoSizeAllColumns' is deprecated and moved to 'api.autoSizeAllColumns'.
> 
> I'm not sure which example is triggering this warning message:
> 
> AG Grid: Grid API function getColumnState() cannot be called as the grid has been destroyed. It is recommended to remove local references to the grid api. Alternatively, check gridApi.isDestroyed() to avoid calling methods against a destroyed grid. To run logic when the grid is about to be destroyed use the gridPreDestroy event

I think I got these working, please check whatever was throwing these warnings.

AnnMarieW:
The [example posted above](https://github.com/plotly/dash-ag-grid/pull/246#issuecomment-1837260467) now works :tada:

Now I'm trying to make it work with various date formats.  For example, if you wanted to create [custom cell data types](https://www.ag-grid.com/javascript-data-grid/cell-data-types/#providing-custom-cell-data-types) and assign it to a column, it doesn't seem to work.  

Note that in the example below, the custom cell data type for the percentage works, but the date does not.


```

import dash_ag_grid as dag
from dash import Dash, html
import pandas as pd


app = Dash(__name__)

df = pd.read_csv(
    "https://raw.githubusercontent.com/plotly/datasets/master/ag-grid/olympic-winners.csv"
)

columnDefs = [
    {"field": 'athlete'},
    {"field": 'age', "cellDataType": "percentage"},
    {"field": 'date', "cellDataType": "dateString2"},
]
defaultColDef = {
      "filter": True,
      "floatingFilter": True,
      "editable": True,
}


app.layout = html.Div(
    [
        dag.AgGrid(
            id="grid-cell-data-types",
            columnDefs=columnDefs,
            defaultColDef=defaultColDef,
            dashGridOptions={"dataTypeDefinitions": {'function': 'dataTypeDefinitions'}},
        ),
    ],
)

app.run(debug=True, port=5000)

"""

dagfuncs.dataTypeDefinitions = {
      dateString2: {
        baseDataType: 'dateString',
        extendsDataType: 'dateString',
        valueParser: (params) =>
          params.newValue != null &&
          params.newValue.match('\\d{2}/\\d{2}/\\d{4}')
            ? params.newValue
            : null,
        valueFormatter: (params) => (params.value == null ? '' : params.value),
        dataTypeMatcher: (value) =>
          typeof value === 'string' && !!value.match('\\d{2}/\\d{2}/\\d{4}'),
        dateParser: (value) => {
          if (value == null || value === '') {
            return undefined;
          }
          const dateParts = value.split('/');
          return dateParts.length === 3
            ? new Date(
                parseInt(dateParts[2]),
                parseInt(dateParts[1]) - 1,
                parseInt(dateParts[0])
              )
            : undefined;
        },
        dateFormatter: (value) => {
          if (value == null) {
            return undefined;
          }
          const date = String(value.getDate());
          const month = String(value.getMonth() + 1);
          return `${date.length === 1 ? '0' + date : date}/${
            month.length === 1 ? '0' + month : month
          }/${value.getFullYear()}`;
        },
      },
      percentage: {
            extendsDataType: 'number',
            baseDataType: 'number',
            valueFormatter: params => params.value == null
                ? ''
                : `${Math.round(params.value * 100)}%`,
        }
    };





"""




```

AnnMarieW:
Could you please add:
```
import 'ag-grid-community/styles/ag-theme-quartz.css';
```
https://github.com/plotly/dash-ag-grid/blob/main/src/lib/fragments/AgGrid.react.js#L54

src/lib/components/AgGrid.react.js
```diff
@@ -770,6 +772,4 @@ const _getAsync =
         );
     };
 export const getApi = _get('getApi');
-export const getColumnApi = _get('getColumnApi');
```
alexcjohnson, line 773: So `getColumnApi` functionality was all merged into `getApi` (and similar for async)? Given that we're exporting these, should we mention this in the [migration guide](https://github.com/plotly/ddk-dash-docs/pull/1938/files#diff-07b135ac5c223e868fdf5b83660e76d692f891059899a28d89b739e9426c7916)?

AnnMarieW: We could mention it in the migration guide, but we haven't yet documented how to use these functions, so it might be a little confusing.

src/lib/fragments/AgGrid.react.js
```diff
@@ -386,6 +388,25 @@ export default class DashAgGrid extends Component {
...
+                }
+                return mapObjIndexed((v) => {
+                    if (typeof v === 'object') {
+                        if ('function' in v) {
+                            if (typeof v.function === 'string') {
+                                return this.convertMaybeFunctionNoParams(v);
+                            }
+                        } else {
+                            return this.convertCol(v);
+                        }
```
alexcjohnson, line 405:
```suggestion
                        if (typeof v.function === 'string') {
                            return this.convertMaybeFunctionNoParams(v);
                        }
                        return this.convertCol(v);
```
I think a similar simplification _can_ happen in all of these blocks, but it's important here so that just having `function` in the object isn't enough to force it into that branch, so that you can still use `function` as the name of a custom data type if you want to.

I debated `typeof v.function` vs `typeof v?.function`), but in this particular case since you already have `if (typeof v === 'object')` around the whole thing the `?` is unnecessary. It brings up another point though that some of these blocks implicitly assume `v` or `value` is an object already, which may or may not be the case - users can supply any sort of crap they want in `dashGridOptions` since we don't have prop type validation there. Might be worth investigating what happens when users provide improper data types, and whether there's anything we can do to make it easier for them to figure out how to fix it.

AnnMarieW: Here's some results of the investigation:
```
 dashGridOptions={"dataTypeDefinitions": "crap"},
```
results in:

image

alexcjohnson: yeah ok - looking at the existing code in more detail, there are all sorts of places we assume the expected data type. So ignore that part of my comment for now, but if we find we're fielding lots of questions about data types we could add some more explicit errors into the converter functions. But my suggested change is still relevant - in fact we need the same down in `convertOne`, line 433.

src/lib/fragments/AgGrid.react.js
```diff
@@ -386,6 +388,25 @@ export default class DashAgGrid extends Component {
                     return c;
                 });
             }
+            if (OBJ_MAYBE_FUNCTION_OR_MAP_MAYBE_FUNCTIONS[target]) {
+                if ('function' in value) {
+                    if (typeof value.function === 'string') {
+                        return this.convertMaybeFunctionNoParams(value);
+                    }
+                }
+                return mapObjIndexed((v) => {
```
alexcjohnson, line 397:
```suggestion
                return map((v) => {
```
minor point, but no need to use the `indexed` form if you're only using the values, not the keys

tests/assets/dashAgGridFunctions.js
```diff
@@ -355,3 +355,56 @@ dagfuncs.contextTest = (params) => {
...
+    const name = cellValue ? cellValue.split(" ")[1] : ""
+    return name && name.toLowerCase().indexOf(filterValues.toLowerCase()) === 0
+}
+// END test_custom_filter.py
+
+// FOR test_quick_filter.py
+dagfuncs.quickFilterMatcher = (quickFilterParts, rowQuickFilterAggregateText) => {
+    return quickFilterParts.every(part => rowQuickFilterAggregateText.match(part));
+}
+// END test_custom_filter.py
```
alexcjohnson, line 410:
```suggestion
// END test_quick_filter.py
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "dash-ag-grid",
-    "version": "2.4.0",
+    "version": "31.0.1rc3",
```
alexcjohnson, line 3: I'd prefer we not try to match the minor or patch version to AG Grid, just the major. That way we still have the flexibility to add features and fix bugs on our own schedule, we just have to hold breaking changes to align with upgrading the AG Grid major version. So the new release should be `31.0.0`

AnnMarieW: @sdidier-dev - can you chime in here?  This may affect how we link to the correct AG Grid version in the docs

sdidier-dev: Yes, I think we need to be in line about how we'll handle the dag version vs AG Grid version to know how to deal with the links to AG Grid docs: if they go to the last version of AG Grid docs or to an "archive" version of their docs.

sdidier-dev: Actually that shouldn't be an issue for the links!
We took a look with Ann Marie, we're using the version of the ag grid package used in dag:
`AG_grid_version = dag.package['dependencies']['ag-grid-community'].lstrip('^')`
then compare it to the latest version of AG Grid coming from their github repo:
`latest_tag_url = "https://api.github.com/repos/ag-grid/ag-grid/releases/latest"`

So the version of dag can be anything that's should be fine for the links to AG Grid docs 🌈

BSd3v: @alexcjohnson,

What are your thoughts on being able to query the underlying grid version from the clientside? Or... could this just be done by searching the source code?

AnnMarieW: It would be handy have a utility -- maybe something like `dag.get_grid_version()`

src/lib/components/AgGrid.react.js
AnnMarieW: Updated [Migration Guide](https://github.com/plotly/ddk-dash-docs/pull/1938/commits/08d0c67537fe3117b57e7b35235e1e751ff4bfe6)  Also updated the notes to add docs for `getApi`

dash_ag_grid/__init__.py
```diff
@@ -23,6 +23,7 @@
 
 package_name = package['name'].replace(' ', '_').replace('-', '_')
 __version__ = package['version']
+get_grid_version = package['dependencies']['ag-grid-community'].lstrip('^')
```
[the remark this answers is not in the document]
AnnMarieW: I agree we should remove the `^`  maybe include a comment so that someone doesn't change it back?

alexcjohnson (approved):
💃 Let's do it!!

PR #247, opened 2023-10-17
JS deps update 2023-10-16

alexcjohnson:
No update to ag-grid (because we'll update that in a #246 and give this package a major version bump)
or to react-markdown (because later versions require React 18)
or react-docgen (because that's used by Dash and IDK what a major version bump will do there)

Everything else is up to date as of yesterday.


alexcjohnson:
hmm `mc001` doesn't like these updates - I guess probably the `remark-gfm` bump from v3 to v4... I'll try downgrading and we can come back to this later.

PR #253, opened 2023-10-26
redirect pythonanywhere docs closes

### LINKED ISSUE #252: Retirement plan for pythonanywhere docs?
Is it time to take the next step toward retiring the pythonanywhere docs? @danton267 pointed out to me that they're still on par with the official docs in Google searches. So they're presumably still getting a good deal of traffic. The banner telling people to switch is great, but it's still an extra step and doesn't take people to the specific page they want.

So it seems to me either we want to turn this site into a page-by-page redirect to the closest analogous page on the official docs, or if we want to keep these docs up, at least update the banner with page-by-page links.

@AnnMarieW thoughts?

AnnMarieW:
Closes #252

### DIFF
--- a/docs_redirects/app.py
+++ b/docs_redirects/app.py
@@ -121,7 +121,7 @@ app.layout =  html.Div("no layout")
 @app.server.before_request
 def redirect():
     print(flask.request.path)
-    return flask.redirect(redirect_map.get(flask.request.path))
+    return flask.redirect(redirect_map.get(flask.request.path, base))
 
 
 if __name__ == "__main__":
--- a/docs_redirects/app.py
+++ b/docs_redirects/app.py
@@ -120,7 +120,6 @@ app.layout =  html.Div("no layout")
 
 @app.server.before_request
 def redirect():
-    print(flask.request.path)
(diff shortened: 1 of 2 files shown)


docs_redirects/app.py
```diff
@@ -0,0 +1,128 @@
...
+}
+
+app = Dash(__name__)
+
+app.layout =  html.Div("no layout")
+
+@app.server.before_request
+def redirect():
+    print(flask.request.path)
+    return flask.redirect(redirect_map.get(flask.request.path))
```
alexcjohnson, line 124:
```suggestion
    return flask.redirect(redirect_map.get(flask.request.path, base))
```
In case there's a link out there that doesn't match one of these pages exactly (try visiting https://dashaggrid.pythonanywhere.com/doesnotexist right now - you get an infinite redirect loop to https://dashaggrid.pythonanywhere.com/None)

alexcjohnson (approved):
💃 Thanks @AnnMarieW!

AnnMarieW:
@alexcjohnson 
It's still not quite right -- if the page doesn't exist, it redirects to the base url, but it looks like this:

image

PR #259, opened 2023-12-13
Add columnsGroupState prop

Lxstr:
Add prop and persistence for column group state in the same fashion and column state


BSd3v:
Hello @Lxstr,

This should be captured in this adjustment and roll out to v31 here:

https://www.ag-grid.com/react-data-grid/grid-state/

AG Grid designed this to completely replicate all of the grid settings without having to recreate them through individual props.

alexcjohnson:
Is this obsolete now that v31 is out?

Lxstr:
Should be, haven't had a chance to confirm yet. I will close as soon as I do :)

PR #261, opened 2023-12-29
Propagate muti-cell paste event to Dash

emilhe:
Currently, the `cellValueChanged` events are propagated to Dash one-by-one. This becomes a problem in cases, where events are fires faster than Dash can handle them, e.g. when pasting multiple cells into the grid. For these cases, only the first event triggers a Dash callback, with all other changes lost. 

This PR changes the event collection mechanism to include all events fired within a specific time window (with the length currently defined by `CELL_VALUE_CHANGED_DEBOUNCE_MS = 1`). By making the window large enough that all events in e.g. a multi cell paste operation can be processed (initial tests indicate that 1 ms is sufficient), all changes can be propagted to Dash. Hence the callback will now receieve a _list_ of change events rather than a _single_ change event.

Here is a small example that demonstrates the usage,

```python
import pandas as pd
import dash_ag_grid as dag
from dash import Dash, Input, Output, html, callback

df = pd.read_csv(
    "https://raw.githubusercontent.com/plotly/datasets/master/ag-grid/olympic-winners.csv"
)

app = Dash(__name__)
app.layout = html.Div(
    [
        dag.AgGrid(
            id="grid",
            columnDefs=[{"field": i} for i in df.columns],
            defaultColDef={"resizable": True, "sortable": True, "filter": True, "editable": True},
            dashGridOptions={
                "enableRangeSelection": True,
            },
            enableEnterpriseModules=True,
            columnSize="sizeToFit",
        ),
        html.Div(id="dummy"),
    ],
)


@callback(
    Output("dummy", "children"),
    Input("grid", "cellValueChanged")
)
def update(cell_changed):
    return str(cell_changed)


if __name__ == "__main__":
    app.run(debug=True)
```

Copy-pasting (the whole) row 0 to row 4 will yield the following output with the main branch,

```bash
{'rowIndex': 4, 'rowId': '4', 'data': {'athlete': 'Michael Phelps', 'age': '23', 'country': 'United States', 'year': '2008', 'date': '24/08/2008', 'sport': 'Swimming', 'gold': '8', 'silver': '0', 'bronze': '0', 'total': '8'}, 'oldValue': 6, 'value': '8', 'colId': 'total', 'timestamp': 1703846436733}
```

With the updates propose in the PR, the output will be,

```bash
[{'rowIndex': 4, 'rowId': '4', 'data': {'athlete': 'Michael Phelps', 'age': '23', 'country': 'United States', 'year': '2008', 'date': '24/08/2008', 'sport': 'Swimming', 'gold': '8', 'silver': '0', 'bronze': '0', 'total': '8'}, 'oldValue': 'Aleksey Nemov', 'value': 'Michael Phelps', 'colId': 'athlete', 'timestamp': 1703846303490}, {'rowIndex': 4, 'rowId': '4', 'data': {'athlete': 'Michael Phelps', 'age': '23', 'country': 'United States', 'year': '2008', 'date': '24/08/2008', 'sport': 'Swimming', 'gold': '8', 'silver': '0', 'bronze': '0', 'total': '8'}, 'oldValue': 24, 'value': '23', 'colId': 'age', 'timestamp': 1703846303491}, {'rowIndex': 4, 'rowId': '4', 'data': {'athlete': 'Michael Phelps', 'age': '23', 'country': 'United States', 'year': '2008', 'date': '24/08/2008', 'sport': 'Swimming', 'gold': '8', 'silver': '0', 'bronze': '0', 'total': '8'}, 'oldValue': 'Russia', 'value': 'United States', 'colId': 'country', 'timestamp': 1703846303491}, {'rowIndex': 4, 'rowId': '4', 'data': {'athlete': 'Michael Phelps', 'age': '23', 'country': 'United States', 'year': '2008', 'date': '24/08/2008', 'sport': 'Swimming', 'gold': '8', 'silver': '0', 'bronze': '0', 'total': '8'}, 'oldValue': 2000, 'value': '2008', 'colId': 'year', 'timestamp': 1703846303491}, {'rowIndex': 4, 'rowId': '4', 'data': {'athlete': 'Michael Phelps', 'age': '23', 'country': 'United States', 'year': '2008', 'date': '24/08/2008', 'sport': 'Swimming', 'gold': '8', 'silver': '0', 'bronze': '0', 'total': '8'}, 'oldValue': '1/10/2000', 'value': '24/08/2008', 'colId': 'date', 'timestamp': 1703846303491}, {'rowIndex': 4, 'rowId': '4', 'data': {'athlete': 'Michael Phelps', 'age': '23', 'country': 'United States', 'year': '2008', 'date': '24/08/2008', 'sport': 'Swimming', 'gold': '8', 'silver': '0', 'bronze': '0', 'total': '8'}, 'oldValue': 'Gymnastics', 'value': 'Swimming', 'colId': 'sport', 'timestamp': 1703846303491}, {'rowIndex': 4, 'rowId': '4', 'data': {'athlete': 'Michael Phelps', 'age': '23', 'country': 'United States', 'year': '2008', 'date': '24/08/2008', 'sport': 'Swimming', 'gold': '8', 'silver': '0', 'bronze': '0', 'total': '8'}, 'oldValue': 2, 'value': '8', 'colId': 'gold', 'timestamp': 1703846303491}, {'rowIndex': 4, 'rowId': '4', 'data': {'athlete': 'Michael Phelps', 'age': '23', 'country': 'United States', 'year': '2008', 'date': '24/08/2008', 'sport': 'Swimming', 'gold': '8', 'silver': '0', 'bronze': '0', 'total': '8'}, 'oldValue': 1, 'value': '0', 'colId': 'silver', 'timestamp': 1703846303493}, {'rowIndex': 4, 'rowId': '4', 'data': {'athlete': 'Michael Phelps', 'age': '23', 'country': 'United States', 'year': '2008', 'date': '24/08/2008', 'sport': 'Swimming', 'gold': '8', 'silver': '0', 'bronze': '0', 'total': '8'}, 'oldValue': 3, 'value': '0', 'colId': 'bronze', 'timestamp': 1703846303493}, {'rowIndex': 4, 'rowId': '4', 'data': {'athlete': 'Michael Phelps', 'age': '23', 'country': 'United States', 'year': '2008', 'date': '24/08/2008', 'sport': 'Swimming', 'gold': '8', 'silver': '0', 'bronze': '0', 'total': '8'}, 'oldValue': 6, 'value': '8', 'colId': 'total', 'timestamp': 1703846303493}]
```

Currently, the time window size is hardcoded, but it could also be changed to be a property of the grid so that users get more fine-grained control. Please let me know, if this approach is preferred or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to `dash-ag-grid` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source Dash AG Grid repo
 
+## UNRELEASED
+
+### Fixed 
+- [#262](https://github.com/plotly/dash-ag-grid/issues/262) Fixed all-but-one `cellValueChanged` events suppressed for multi-cell edits
+
 ## [2.4.0] - 2023-10-17
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,9 @@ Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source D
 
 ## UNRELEASED
 
-### Fixed 
-- [#262](https://github.com/plotly/dash-ag-grid/issues/262) Fixed all-but-one `cellValueChanged` events suppressed for multi-cell edits
+### Changed 
+
+- The `cellValueChanged` property has changed been changed from a (single) event object to a _list_ of event objects. For multi-cell edits, the list will contain an element per change. In other cases, the list will contain a single element. Fixes [#262](https://github.com/plotly/dash-ag-grid/issues/262) 
 
 ## [2.4.0] - 2023-10-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source D
 
 ### Changed 
 
-- The `cellValueChanged` property has changed been changed from a (single) event object to a _list_ of event objects. For multi-cell edits, the list will contain an element per change. In other cases, the list will contain a single element. Fixes [#262](https://github.com/plotly/dash-ag-grid/issues/262) 
+- [#261](https://github.com/plotly/dash-ag-grid/pull/261) The `cellValueChanged` property has changed been changed from a (single) event object to a _list_ of event objects. For multi-cell edits, the list will contain an element per change. In other cases, the list will contain a single element. Fixes [#262](https://github.com/plotly/dash-ag-grid/issues/262) 
 
 ## [2.4.0] - 2023-10-17
 
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -925,22 +925,22 @@ export default class DashAgGrid extends Component {
             value,
             colId,
             timestamp,
-        }
+        };
         // Append it to current change session.
         let pendingChanges = this.state.cellValueChanged;
-        if(pendingChanges === undefined){
-            pendingChanges = [newChange]
+        if(typeof pendingChanges === 'undefined' || pendingChanges === null){
+            pendingChanges = [newChange];
         }
         else{
-            pendingChanges.push(newChange)
+            pendingChanges.push(newChange);
         }
-        this.setState({cellValueChanged: pendingChanges})
+        this.setState({cellValueChanged: pendingChanges});
     }
 
     afterCellValueChanged() {
-        const cellValueChanged = this.state.cellValueChanged;
+        const {cellValueChanged} = this.state;
         // Guard against multiple invocations of the same change session.
-        if(cellValueChanged === undefined){
+        if(typeof cellValueChanged === 'undefined' || cellValueChanged === null){
             return;
         }
         // Send update(s) for current change session to Dash.
@@ -951,7 +951,7 @@ export default class DashAgGrid extends Component {
         });
         this.syncRowData();
         // Mark current change session as ended.
-        this.setState({cellValueChanged: undefined})
+        this.setState({cellValueChanged: null});
     }
 
     onDisplayedColumnsChanged() {
@@ -1353,7 +1353,7 @@ export default class DashAgGrid extends Component {
                     onCellDoubleClicked={this.onCellDoubleClicked}
                     onCellValueChanged={debounce(
                         this.afterCellValueChanged,
-                        CELL_VALUE_CHANGED_DEBOUNCE_MS, // FIXME: SET VARIABLE(!)
+                        CELL_VALUE_CHANGED_DEBOUNCE_MS,
                         this.onCellValueChanged,
                     )}
                     onFilterChanged={this.onFilterChanged}
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -185,6 +185,7 @@ export default class DashAgGrid extends Component {
 
         this.selectionEventFired = false;
         this.reference = React.createRef();
+        this.pendingChanges = null;
     }
 
     onPaginationChanged() {
@@ -927,13 +928,12 @@ export default class DashAgGrid extends Component {
             timestamp,
         };
         // Append it to current change session.
-        let pendingChanges = this.state.cellValueChanged;
-        if (typeof pendingChanges === 'undefined' || pendingChanges === null) {
-            pendingChanges = [newChange];
+        if (typeof this.pendingChanges === 'undefined' || this.pendingChanges === null) {
+            this.pendingChanges = [newChange];
         } else {
-            pendingChanges.push(newChange);
+            this.pendingChanges.push(newChange);
         }
-        this.setState({cellValueChanged: pendingChanges});
+        this.setState({cellValueChanged: this.pendingChanges});
     }
 
     afterCellValueChanged() {
@@ -954,6 +954,7 @@ export default class DashAgGrid extends Component {
         this.syncRowData();
         // Mark current change session as ended.
         this.setState({cellValueChanged: null});
+        this.pendingChanges = null;
     }
 
     onDisplayedColumnsChanged() {
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -113,10 +113,6 @@ function stringifyId(id) {
     return '{' + parts.join(',') + '}';
 }
 
-function is_undefined_or_null(obj) {
-    return typeof obj === 'undefined' || obj === null;
-}
-
 export default class DashAgGrid extends Component {
     constructor(props) {
         super(props);
@@ -932,7 +928,7 @@ export default class DashAgGrid extends Component {
             timestamp,
         };
         // Append it to current change session.
-        if (is_undefined_or_null(this.pendingCellValueChanges)) {
+        if (!this.pendingCellValueChanges) {
             this.pendingCellValueChanges = [newChange];
         } else {
             this.pendingCellValueChanges.push(newChange);
@@ -941,7 +937,7 @@ export default class DashAgGrid extends Component {
 
     afterCellValueChanged() {
         // Guard against multiple invocations of the same change session.
-        if (is_undefined_or_null(this.pendingCellValueChanges)) {
+        if (!this.pendingCellValueChanges) {
             return;
         }
         // Send update(s) for current change session to Dash.
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -66,7 +66,7 @@ const RESIZE_DEBOUNCE_MS = 200;
 const COL_RESIZE_DEBOUNCE_MS = 500;
 
 // Time between syncing cell value changes with Dash
-const CELL_VALUE_CHANGED_DEBOUNCE_MS = 10;
+const CELL_VALUE_CHANGED_DEBOUNCE_MS = 100;
 
 const xssMessage = (context) => {
     console.error(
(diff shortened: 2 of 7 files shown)


BSd3v:
Have you tested out what this does with row editing and changing all three values?

BSd3v:
Yes, with fullRowEditing. 

There is a test that uses this functionality, it adds a class to the cell based upon if it was edited. 

Increase it to editing more than one column before moving on and see if this catches the event properly.

emilhe:
@BSd3v do you have a link to a test/example that demonstrates/uses this functionality (in Dash)?

emilhe:
@BSd3v thanks! For `fullRowEditing`, only a single event is captured at a time. I think that makes sense; event though the whole row is being edited, only a single cell is edited at the time, so there should be one event per callback.

BSd3v:
Even when you have edited multiple?

emilhe:
@BSd3v yes. The event 'bundling' mechanism is time-based, so it doesn't matter that you are in this particular mode.

BSd3v:
Here, run this, I get three when I edit 3 cells and then move down to the next row:

```
import dash_ag_grid as dag
from dash import *
import pandas as pd


app = Dash(__name__)

df = pd.read_csv(
    "https://raw.githubusercontent.com/plotly/datasets/master/ag-grid/olympic-winners.csv"
)

app.layout = html.Div(
    [
        dag.AgGrid(
            columnDefs=[{'field': x} for x in df.columns],
            columnSize="sizeToFit",
            defaultColDef={"editable": True},
            id="grid",
            getRowId="params.data.nation",
            dashGridOptions={'editType': 'fullRow'}
        ),
        html.Div(id="log")
    ]
)

app.clientside_callback(
    """function countEvents(changes) {
        console.log("FIRE");
        return changes? changes.length : 0;
    }""",
    Output("log", "children"),
    Input("grid", "cellValueChanged"),
    prevent_initial_call=True,
)

app.run(debug=True, port=5000)
```

This is what I was wondering about.

BSd3v:
I'm merging into v31 bump, its possible that this is able to handle correctly in that version with React 18.

emilhe:
@BSd3v I dont really understand what the issue is with React 18 - but if it's solved, that would be great 😃

BSd3v:
@emilhe, me neither.

You can run app's in React 18 like this:
```
external_scripts=[
               'https://cdnjs.cloudflare.com/ajax/libs/react/18.2.0/umd/react.production.min.js',
               'https://cdnjs.cloudflare.com/ajax/libs/react-dom/18.2.0/umd/react-dom.production.min.js'
           ]
```

BSd3v:
Looks like in React18, the state isnt even updated when it triggers the second time

emilhe:
@BSd3v Thanks! I'll try that out. Do you know how to clear multiple cells at once from the test (like when you select using click + shift and click delete)?

BSd3v:
You can only do it with range selection, you cant do it without enterprise. 😛

BSd3v:
Eh, you can do this:

```
import dash

dash._dash_renderer._set_react_version("18.2.0")
```

To run in v18.

I have a version working now, it just updates the variable instead of the state.

emilhe:
Thanks for the update! I have adjusted/cleanup up the code a bit.

BSd3v:
> Thanks for the update! I have adjusted/cleanup up the code a bit. I also think that I understand why the initial version had problems with React 18; version 18 introduces some performance [optimizations](https://www.freecodecamp.org/news/react-18-new-features/), which may introduce race conditions, which could explain the behavior we saw.

I was thinking that was along the lines of it, haha. Much more performant means things that were in order could not be, haha.

emilhe:
@BSd3v about the `fullRowEditing` case. You are correct - the events fires only when the row editing finishes. Hence, with the current implementation, all but a single edit event is lost. With the new implementation, all of the events are collected. So this PR also solves this issue for that case.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to `dash-ag-grid` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source Dash AG Grid repo
 
+## UNRELEASED
+
+### Fixed 
+- [#262](https://github.com/plotly/dash-ag-grid/issues/262) Fixed all-but-one `cellValueChanged` events suppressed for multi-cell edits
```
alexcjohnson, line 10: Rather than `### Fixed` this should be a `### Changed` with a very prominent note that it's a breaking change. So it's good that we're about to release a new major version for AG Grid v31 or this kind of change would not be allowed 😅 But it makes sense, I don't see any non-breaking way to support these events.

@BSd3v this makes me wonder, are there any other props that might have similar problems? If so now is the time to convert them into arrays too!

src/lib/fragments/AgGrid.react.js
```diff
@@ -65,6 +65,9 @@ const RESIZE_DEBOUNCE_MS = 200;
 // Rate-limit for updating columnState when interacting with the grid
 const COL_RESIZE_DEBOUNCE_MS = 500;
 
+// Time between syncing cell value changes with Dash
+const CELL_VALUE_CHANGED_DEBOUNCE_MS = 1;
```
alexcjohnson, line 69: 0 might even be sufficient here, these events are likely coming in from sync code, if so all you need to do is wait for the next tick in the event loop. But that's just FYI, let's leave it as 1, there could be other problems with 0.

alexcjohnson:
@emilhe this looks great, thanks! I just had 2 small comments, then will let @BSd3v merge as he sees fit in conjunction with the v31 work.

CHANGELOG.md
emilhe: If backwards compatibility is a main concern, we could

* Set the type of `cellValueChanged ` to union of single/list of events
* Add a new prop that represents `CELL_VALUE_CHANGED_DEBOUNCE_MS `, which defaults to -1, in which case we fallback to the old behavior, thus returning a single element. Setting the value it to 0 (or higher) will enable the new behavior, thus returning a list 

My initial implementation was actually along these lines, but @BSd3v suggested to target a simpler implementation (the one in this PR) in favor of backwards compatibility since a major release is in the making.

src/lib/fragments/AgGrid.react.js
emilhe: I think you are right! I played around with different values, but i didn't seem to matter if it was 0, 1 or 100.

CHANGELOG.md
alexcjohnson: Yep, I agree the simpler API is more important than backward compatibility, given the upcoming major release and the fact it would be annoying to use the prop if sometimes it’s a list and sometimes not. Just make sure the changelog makes the breaking change clear.

src/lib/fragments/AgGrid.react.js
```diff
@@ -912,20 +921,38 @@ export default class DashAgGrid extends Component {
...
+            rowIndex,
+            rowId: node.id,
+            data,
+            oldValue,
+            value,
+            colId,
+            timestamp,
+        };
+        // Append it to current change session.
+        if (is_undefined_or_null(this.pendingCellValueChanges)) {
```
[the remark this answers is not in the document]
BSd3v: @emilhe, what are your thoughts on this? I'm curious if there is a performance pick up, not that it needs it, but still curious.

BSd3v (commented):
Nothing really to add, just those two things.

src/lib/fragments/AgGrid.react.js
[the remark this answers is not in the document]
emilhe: I wrote the code initially to be as explicit and robust as possible, hence the explicit checks. But I believe that Alexs code will be equvalent as the code currently stands (there would be differences if `pendingChanged` toke a value if `0` for example, but it doesn't in the current implementation). In terms of performance there should be no difference at all. In favor of simplicity, I'll adopt Alexs suggestion for now.

CHANGELOG.md
emilhe: I have updated the comment, indicating that this is a _change_. Should I note explicitly that the change is breaking (I am not sure how breaking changes are documented in this project - I didn't see any mentioned of breaking changes previously in the changelog)?

BSd3v: This would be the first since switching to version 2 from 1.

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to `dash-ag-grid` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source Dash AG Grid repo
 
+## UNRELEASED
+
+### Changed 
+
+- The `cellValueChanged` property has changed been changed from a (single) event object to a _list_ of event objects. For multi-cell edits, the list will contain an element per change. In other cases, the list will contain a single element. Fixes [#262](https://github.com/plotly/dash-ag-grid/issues/262)
```
BSd3v, line 11:
```suggestion
- [#261](https://github.com/plotly/dash-ag-grid/pull/261) The `cellValueChanged` property has changed been changed from a (single) event object to a _list_ of event objects. For multi-cell edits, the list will contain an element per change. In other cases, the list will contain a single element. Fixes [#262](https://github.com/plotly/dash-ag-grid/issues/262) 
```

This is our standard for making the changes, so that they point to a specific PR.

BSd3v (changes requested):
💃 after the slight modification to the changelog

BSd3v (approved):
💃

BSd3v (approved):

olejorgenb:
Will two or more edits not originating from paste events, multi delete or fullRowEditing (or similar) realistically ever be batched?

emer-bestseller:
What other cases are you thinking about?

olejorgenb:
I wonder if there's cases where I can be sure list will only contain one change, since the logic is simpler to implement (or rather - I have existing logic that is a bit hairy to refactor). In my case don't think it's possible to paste content for instance.

Ie.: will a user be able to trigger grouped changes by changing multiple cells manually fast?

PR #266, opened 2024-02-03
Update readme

AnnMarieW:
Updated the README.md


README.md
```diff
@@ -43,25 +123,3 @@ Now install the component in development mode:
 $ pip install -e .
 ```
 In development mode, Python uses the files in this directory when you import the package. So you can write a testing app in another folder, and whenever you change some code and rebuild the component here it will update in your testing app.
-
-### Create a production build
```
alexcjohnson, line 47: Perhaps we can start a `CONTRIBUTING.md` and move this section and the `Contributing` and `Developing in this repo` sections into it? I like to have the publishing instructions included in the repo somewhere, but you're right that they don't belong in the main `README.md`.

PR #270, opened 2024-02-13
Event Listeners

BSd3v:
allowing the grid to be passed `eventListeners` that will be added upon `gridReady`, these events will also have access to `setGridProps` which can interact with the grid info directly.


src/lib/components/AgGrid.react.js
```diff
@@ -526,6 +527,12 @@ DashAgGrid.propTypes = {
         columnPosition: PropTypes.oneOf(['auto', 'start', 'middle', 'end']),
     }),
 
+    /**
+     * Object of Eventlisteners to add upon grid ready. These listeners are only added upon grid ready,
+     * to add or remove an event listener after this point, please utilize the `getApi` or `getApiAsync` methods.
```
alexcjohnson, line 532:
```suggestion
     * Object of Eventlisteners to add upon grid ready. These listeners are only added upon grid ready.
     * To add or remove an event listener after this point, please utilize the `getApi` or `getApiAsync` methods.
```

alexcjohnson (approved):
💃 LGTM! Just a minor grammatical suggestion

PR #273, opened 2024-02-28
Update getApiAsync to allow for a longer timeout

AnnMarieW:
Allow for up to 2 minutes before `getApiAsync` fails.


alexcjohnson (approved):
💃 Looks great!

Issue #282, opened 2024-04-07
Automatically select the first row


brifordwylie:
So my use case is probably not that rare, you have a table of items at the top of the page and then details about those items in other components. So when the item table (ag-grid) gets populated you'd like the first row selected, and therefore a "selectedRows" event/callback gets invoked. That will populate the bottom components with the details of the selected row.  Here's an example. The first row is selected 'automatically' and therefore populates the bottom components. 

<img width="500" alt="Screenshot 2023-12-29 at 8 38 42 PM" src="https://github.com/plotly/dash-ag-grid/assets/4806709/3681a1d9-8230-4215-be1b-3546db4d443f">

When using dash DataTable, this is done via the 'selected_rows' attribute...
```
 table = dash_table.DataTable(
            id=component_id,
            columns=[{"name": "Status", "id": "status"}],
            data=[{"status": "Waiting for data..."}],
            sort_action="native",
            row_selectable=row_select,
            cell_selectable=False,
            selected_rows=[0],
```

I'm not finding the equivalent in Ag-grid, my stack overflow/chat-gpt is failing me :)

Here's a reproducible example.. any suggestions are greatly appreciated...
```
import dash
import pandas as pd
from dash_ag_grid import AgGrid

app = dash.Dash(__name__)

def select_first_row_js():
    """Return the JavaScript to select the first row in the table"""
    return (
        "function(params) {"
        "  var firstRowNode = params.api.getDisplayedRowAtIndex(0);"
        "  if (firstRowNode) {"
        "    firstRowNode.setSelected(true);"
        "  }"
        "}"
    )

df = pd.DataFrame({
    "A": [1, 2, 3],
    "B": [4, 5, 6]
})

app.layout = AgGrid(
    id='my-table',
    columnSize="sizeToFit",
    dashGridOptions={
        "rowHeight": None,
        "domLayout": "normal",
        "rowSelection": "single",
        "filter": True,
        "firstDataRendered": select_first_row_js(),
    },
    style={"maxHeight": "200px", "overflow": "auto"},
    columnDefs=[{"headerName": col, "field": col, "filter": "agTextColumnFilter"} for col in df.columns],
    rowData=df.to_dict("records")
)

if __name__ == '__main__':
    app.run_server(debug=True)
```

AnnMarieW:
Hi @brifordwylie 

Try removing `"firstDataRendered": select_first_row_js(),` and using the `selectedRows` prop
```python

app.layout = AgGrid(
[...]

PR #281, opened 2024-04-04
adjustments to webpack to improve build speed

BSd3v:
adjustments to webpack to improve build speed

Issue #274, opened 2024-02-28
Several issues with selectedRows as an output


bhofmei:
I am encountering numerous issues with `selectedRows` in v31.0.1 when used as an output that were not problems in v2.4. Since there are numerous issues to illustrate, I'll provide the set-up and layout at the beginning then updated callbacks for each. Examples are based on the simple demo in the documents.

```python
import random, string
import dash_ag_grid as dag
from dash import Dash, html, Input, Output, callback, dcc, ctx
import pandas as pd

app = Dash(__name__)
def random_string(length):
    """Return a random string of length"""
    return ''.join(random.choices(string.ascii_uppercase + string.ascii_lowercase + string.digits, k=length))

df = pd.read_csv(
    "https://raw.githubusercontent.com/plotly/datasets/master/ag-grid/olympic-winners.csv"
)

columnDefs = [{'field': 'country', 'checkboxSelection': True, 'headerCheckboxSelection': True}] + [{"field": i} for i in ["year", "athlete", "age", "sport", "total"]]

app.layout = html.Div(
    [
        html.Div('Athletes', id='header'),
        html.Button('Pull data', id='pull-data-btn'),
        dag.AgGrid(
            id="row-selection-selected-rows",
            rowData=[],
            columnDefs=columnDefs,
            getRowId='params.data.uuid',
            dashGridOptions={"rowSelection": "multiple", 'suppressRowClickSelection': True},
        ),
        html.Pre(id="pre-row-selection-selected-rows", style={'text-wrap': 'wrap'}),
        html.Pre(id='row-ids', style={'text-wrap': 'wrap'}),
        dcc.Store(id='data-store')
    ],
)
```

1. When using rowIds and checkbox selection, the checkboxes are never visibly selected.
```python
@callback(
    Output("pre-row-selection-selected-rows", "children"),
    Input("row-selection-selected-rows", "selectedRows"),
)
def output_selected_rows(selected_rows):
    if selected_rows is None:
        return "No selections"
    selected_list = [f"{s['athlete']} ({s['year']})" for s in selected_rows]
    return f"You selected the athlete{'s' if len(selected_rows) > 1 else ''}:\n{', '.join(selected_list)}" if selected_rows else "No selections"

@callback(
    Output("row-selection-selected-rows", "selectedRows"),
    Output('row-ids', 'children'),
    Input("data-store", "data"),
)
def select_rows(data):
    if not data or data == {}:
        return [], 'No data'
[...]

PR #283, opened 2024-04-09
Selected rows fixes

BSd3v:
- adding timeouts to keep async processes together

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  browser-tools: circleci/browser-tools@1.4.4
+  browser-tools: circleci/browser-tools@1.4.8
 
 jobs:
   test: &test
@@ -12,9 +12,8 @@ jobs:
           password: $DASH_PAT_DOCKERHUB
     steps:
       - checkout
-      - browser-tools/install-browser-tools:
-          chrome-version: 116.0.5845.96
-          firefox-version: 116.0.3
+      - browser-tools/install-chrome
+      - browser-tools/install-chromedriver
       - run:
           name: Install Python deps
           command: |
@@ -36,6 +35,8 @@ jobs:
           name: Run tests
           command: |
             . venv/bin/activate
+            pip uninstall -y pytest
+            pip install 'pytest<8.1.0'
             pytest --headless
   test-react-18:
     <<: *test
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -176,7 +176,6 @@ export default class DashAgGrid extends Component {
                 markdown: this.generateRenderer(MarkdownRenderer),
                 ...newComponents,
             },
-            pauseSelections: false,
             rerender: 0,
             openGroups: {},
             gridApi: null,
@@ -184,8 +183,10 @@ export default class DashAgGrid extends Component {
         };
 
         this.selectionEventFired = false;
+        this.pauseSelections = false;
         this.reference = React.createRef();
         this.pendingChanges = null;
+        this.dataUpdates = false;
     }
 
     onPaginationChanged() {
@@ -204,11 +205,10 @@ export default class DashAgGrid extends Component {
         }
     }
 
-    setSelection(selection) {
-        const {gridApi} = this.state;
+    setSelection(selection, gridApi = this.state?.gridApi) {
         const {getRowId} = this.props;
         if (gridApi && selection && !gridApi?.isDestroyed()) {
-            this.setState({pauseSelections: true});
+            this.pauseSelections = true;
             const nodeData = [];
             if (has('function', selection)) {
                 const test = this.parseFunction(selection.function);
@@ -254,7 +254,9 @@ export default class DashAgGrid extends Component {
             }
             gridApi.deselectAll();
             gridApi.setNodesSelected({nodes: nodeData, newValue: true});
-            this.setState({pauseSelections: false});
+            setTimeout(() => {
+                this.pauseSelections = false;
+            }, 1);
         }
     }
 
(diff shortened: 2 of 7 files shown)


brifordwylie:
this is looking promising. I tried to check out the branch and test it out (even before merge) just to see if my test case would work. I got an error when importing the  dash_ag_grid module (might be pilot error). Here's the reproduction steps...

```
git clone git@github.com:BSd3v/dash-ag-grid-real.git dash-ag-grid-beta
cd dash-ag-grid-beta
<commit = commit 7acd81ea (HEAD -> selectedRows-fixes, origin/selectedRows-fixes)>
pip install -e .
```
Then just try importing the package..
```
ipython
In [1]: import dash_ag_grid
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[1], line 1
----> 1 import dash_ag_grid

File ~/work/dash-ag-grid-beta/dash_ag_grid/__init__.py:10
      7 import dash as _dash
      9 # noinspection PyUnresolvedReferences
---> 10 from ._imports_ import *
     11 from ._imports_ import __all__
     13 if not hasattr(_dash, '__plotly_dash') and not hasattr(_dash, 'development'):

ModuleNotFoundError: No module named 'dash_ag_grid._imports_'
```

brifordwylie:
Ah, silly me... I think the Contributing.md shows them smooshed together.  Anyway running my test code against your branch looks good... I can callback/update the table with "columnDefs", "rowData", "selectedRows" and it all looks like it working properly. Thanks a bunch for the fix. Looking forward to new PyPI version when you get a chance :)

.circleci/config.yml
```diff
@@ -36,6 +35,8 @@ jobs:
           name: Run tests
           command: |
             . venv/bin/activate
+            pip uninstall -y pytest
+            pip install 'pytest<8.1.0'
```
ndrezn, line 39: Why is there an `uninstall` step here? Also, why pin at <8.1.0?

BSd3v: Pytest has an issue with something and was failing. Reverting back to a lower version solved the issue.

src/lib/fragments/AgGrid.react.js
```diff
@@ -176,16 +176,17 @@ export default class DashAgGrid extends Component {
...
             openGroups: {},
             gridApi: null,
             columnState_push: true,
         };
 
         this.selectionEventFired = false;
+        this.pauseSelections = false;
         this.reference = React.createRef();
         this.pendingChanges = null;
+        this.dataUpdates = false;
```
ndrezn, line 189: What are the new `pauseSelections`/`dataUpdates` props for?

.circleci/config.yml
```diff
@@ -12,9 +12,8 @@ jobs:
           password: $DASH_PAT_DOCKERHUB
     steps:
       - checkout
-      - browser-tools/install-browser-tools:
-          chrome-version: 116.0.5845.96
-          firefox-version: 116.0.3
+      - browser-tools/install-chrome
```
ndrezn, line 15: These diffs are all coming from #281 right? So we'll need to merge #281 first, rebase, then merge in these changes..?

BSd3v: Yes. That is correct.

src/lib/fragments/AgGrid.react.js
BSd3v: `pauseSelections` isnt new, its to keep systematic adjustments from triggering the `onSelectionChanged`, as far as `dataUpdates` its to be used in the event of `rowData` pushed from the server, this way it can ignore selection events. As this clears out the `selectedRows` if there is no data there:

`rowData` became a async process, so both were triggered at the same time. But because the `rowData` was updating, when the `selectedRows` were trying to be applied the grid was essentially empty.

ndrezn: Got it. Just wondering, is there an open ticket that describes these issues? #274 looks like it's describing these issues..?

BSd3v: Yes, this is linked with the two issues already. :)

ndrezn: aha... yep catching up to you hah

tests/test_selected_rows.py
```diff
@@ -95,3 +97,75 @@ def selected(n):
     dash_duo.wait_for_text_to_equal(
         "#selections-single-output", "You selected athlete: Natalie Coughlin"
     )
+
+def test_sr2_selected_rows_rowdata(dash_duo):
```
ndrezn, line 101: Reviewing the test cases in #274:

We're covering:
- Cannot output rowData and selectedRows in the same callback

But confirming this covers:
- Inconsistent data type when selectedRows is an input depending how selectedRows was specified
- When using rowIds and checkbox selection, the checkboxes are never visibly selected

Could you add a comment to this test describing these cases as well so it's clear what it's testing for?

This example is much clearer and more concise overall than the originals, kudos!

tests/test_selected_rows.py
```diff
@@ -95,3 +97,75 @@ def selected(n):
...
+    )
+    def update_data(n_clicks):
+        df = pd.DataFrame({
+            "A": [1, 2, 3],
+            "B": [4, 5, 6]
+        })
+        column_defs = [{"headerName": col, "field": col, "filter": "agTextColumnFilter"} for col in df.columns]
+        row_data = df.to_dict("records")
+        selected_rows = df.head(1).to_dict("records")
+        return column_defs, row_data, selected_rows
```
ndrezn, line 138: Nice test for selecting the first row for #282 👍

ndrezn (approved):
💃 Please rebase from `main` and then I can merge! Nice change and great test case!

BSd3v:
Closed as this had rebasing issues, replaced by:

https://github.com/plotly/dash-ag-grid/pull/290

PR #287, opened 2024-04-22
bump to v31.1.0

BSd3v:


package.json
```diff
@@ -31,9 +31,9 @@
     "dependencies": {
         "@emotion/react": "^11.11.3",
         "@emotion/styled": "^11.11.0",
-        "ag-grid-community": "31.1.1",
-        "ag-grid-enterprise": "31.1.1",
-        "ag-grid-react": "31.1.1",
+        "ag-grid-community": "31.2.1",
+        "ag-grid-enterprise": "31.2.1",
+        "ag-grid-react": "31.2.1",
```
ndrezn, line 36: @BSd3v make sure to commit changes to `package-lock.json` here

BSd3v: AG Grid and our package version should be insync, thus its easier to know what version dag is pointing to.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "dash-ag-grid",
-    "version": "31.0.1",
+    "version": "31.2.1",
```
ndrezn, line 3: I thought our version sync pattern was to match `major.minor` from Ag Grid, and then we reserve `patch` for us. So this bump would be to `31.2.0` and if we need to issue any patch releases without bumping the Ag Grid version _then_ we go to `31.2.1`. But, @alexcjohnson / @BSd3v correct me if I'm wrong.

ndrezn: Basically this is a problem because if we match 1:1 then we're blocked from releasing any changes until Ag Grid makes a new release. Ag Grid releases frequently than us so most of the time this should be okay but food for thought.

BSd3v: I think that would be the case. We can revert this since you have automated the process of creating a draft. ;)

ndrezn: You can also add another commit to this repo just bumping the patch version down to 0; no need to go through the fuss of opening a new PR.

ndrezn: Part of my other comment: Basically the changelog should make clear which version (_including the patch_) of Ag Grid we're bundling -- so in our 31.2.0, we bundle 31.2.1: This isn't obvious and should be marked in the changelog.

ndrezn: I'll update the release process since this all feels kinda arcane 🙃

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "dash-ag-grid",
-    "version": "31.2.1",
+    "version": "31.0.1",
```
ndrezn, line 3: Should be 31.2.0.

BSd3v: I just reverted it, otherwise this will create a draft. 😛

ndrezn: A draft what?

BSd3v: Of the release, right?

ndrezn: The release draft will only be generated when we push a tag

BSd3v: Ah, ok. XD

BSd3v:
I'm working on something that I think will help with the flakey ness of the selectedRows tests

CHANGELOG.md
```diff
@@ -8,13 +8,24 @@ Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source D
 
 ### Changed
```
ndrezn, line 9: There should be a new version number added rather than `unreleased` similar to others

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "dash-ag-grid",
-    "version": "31.0.1",
+    "version": "31.2.0",
```
ndrezn, line 3: Chatted with @alexcjohnson -- I was wrong with the version bump; we follow the _major_ version of Ag Grid only, so this would be a minor bump to `31.1.0`. Sorry for the confusion here...

ndrezn (commented):
Small comments to address, and then 💃 

In the future I think we should make Ag Grid version bumps their own PR (this would also catch the test issue in that PR) but don't want to slow down progress for the sake of procedure 😛

CHANGELOG.md
BSd3v: unreleased swaps to version number with the release of the new package, as we dont know what version these changes will be included in when adding. 😏

ndrezn: This is the release PR though?

PR #288, opened 2024-04-22
Update requires files

ndrezn:
Moves our requirements into a more accessible place.


BSd3v (approved):
💃 

Looks good to me, as long as it works, haha.

ndrezn:
Should be in business! The thing to test is whether the build with:
```
python setup.py sdist bdist_wheel
```

is valid. Also, take note of the new `docs` and `dev` extras patterns which make it faster and more reliable to install dependencies for a specific development environment.

PR #289, opened 2024-04-22
Use Github Actions rather than Circle for automated tests

ndrezn:
Our other projects use GHA - let's switch this over too.


.github/workflows/python-test.yml
```diff
@@ -0,0 +1,96 @@
+name: CI Pipeline
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
```
ndrezn, line 10: This also adds test cases for our full set of actively supported Python versions. Technically we support `>=3.6` but in a future release we will support only `>=3.8`.

.github/workflows/python-test.yml
```diff
@@ -0,0 +1,96 @@
...
+        run: |
+            export DISPLAY=:99
+            chromedriver --url-base=/wd/hub &
+      - name: Start XVFB
+        run: Xvfb :99 &
+
+      - name: Setup uv
+        run: |
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          uv venv
```
ndrezn, line 79: I'm using `uv` instead of `pip` for package installation to improve performance.

.circleci/config.yml
ndrezn, line 1: 🫡

.github/workflows/python-test.yml
```diff
@@ -0,0 +1,102 @@
...
+      - name: Run tests
+        run: |
+          source .venv/bin/activate
+          pytest --headless
+
+  test-react:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        react-version: ["18.2.0"]
```
ndrezn, line 66: @alexcjohnson , why was the previous test running a special check on `18.2.0`?

BSd3v: This was to test for forward compatibility for dash 2.17.

ndrezn: Cool I suppose we can keep it till 2.17 rolls out

BSd3v:
This is only failing due to circleci missing?

ndrezn:
Yeah I need to remove the circleCI config for this project. Then we should be good!

ndrezn:
I paused CircleCI builds on this project so this should pass with Circle on merge.

.github/workflows/python-test.yml
alexcjohnson: Definitely min 3.8 and max 3.12, but I’m not sure there’s much value for components to test everything between, since there’s very little Python code in the component

ndrezn: I'll reduce the matrix to just 3.12 & 3.8

BSd3v (approved):
💃

.circleci/config.yml
BSd3v: We thank you for your service.

PR #291, opened 2024-04-23
Automate generation of build files

ndrezn:
With this PR, when a new tag `v*` is pushed, a draft release will be created with the build files automatically attached.

Then, someone from the Plotly team should publish that draft release and upload the files to PyPI using Twine.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,6 +28,7 @@ jobs:
           source .venv/bin/activate
           npm i
           npm run dist
+          npm pack && mv *.tgz dist/
         shell: bash
       - uses: ncipollo/release-action@v1
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "v18.16.0"
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
```
BSd3v, line 21: does this match the python-version in the AG Grid tests?

ndrezn: The tests run on 3.8-3.12 inclusive: Because of this we expect a consistent build on any version and can increment this with some regular cadence.

ndrezn:
@alexcjohnson if you're curious about the release process we chatted about.

BSd3v (approved):
💃 

As long as it works. 👍

PR #294, opened 2024-04-25
Explain versioning system

ndrezn:
Explains how versioning works, and adds a test to ensure we actually follow the system!


tests/test_versioning.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+    Test that the major and minor versions of bundled Ag Grid align with
+    the major and minor versions of the Python package.
+
+    See CONTRIBUTING.md for more information on versioning this package.
+    """
+    ag_version = parse(dag.grid_version)
+    dash_ag_version = parse(dag.__version__)
+
+    assert (
```
ndrezn, line 15: Hm I guess we only would want to run this test when the version is bumped...

BSd3v: That's why it needs to be a test before dist and not on github

ndrezn: Touche.

tests/test_versioning.py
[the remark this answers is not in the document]
ndrezn: hm yeah, that's a good idea. Let me think about this for a minute. It might make sense to move it out, or to have a tests config file to exclude it unless specified..

gvwilson:
@ndrezn @BSd3v can we please give this some love?

BSd3v:
@ndrezn, @gvwilson, thoughts on the changes?

Should we describe the testing behaviour in the contributing guide?

eg: `test_version.py` => AG-grid in unison within itself, `test_versioning.py` DAG is in unison with AG-grid?

ndrezn:
Yes, definitely worth adding a note to the contribution guide. I'm not sure also if we mention running the tests locally but devs probably should be recommended to use `pytest --headless -k "not test_versioning"` rather than just `pytest`.

BSd3v:
`test_versioning.py` isnt in the `tests/` anymore. So they can just run `pytest`. 😄

CONTRIBUTING.md
```diff
@@ -37,7 +37,7 @@ We may release out-of-band of Ag Grid when there are patches that we want to mak
 
 As a user, you can always check the underlying Ag Grid version with `dash_ag_grid.grid_version` and the underlying Dash Ag Grid version with `dash_ag_grid.__version__`.
 
-For maintainers, when issuing new releases ensure that the version bump of Dash Ag Grid follows this convention. This can be verified after a build by using `npm run pre-flight-dag-version` or `python test_versioning.py`.
+For maintainers, when issuing new releases ensure that the version bump of Dash Ag Grid follows this convention. This can be verified after a build by using `npm run pre-flight-dag-version` or `python test_versioning.py`. This is automatically done during the `npm run dist`
```
ndrezn, line 40:
```suggestion
For maintainers, when issuing new releases ensure that the version bump of Dash Ag Grid follows this convention. This can be verified after a build by using `npm run pre-flight-dag-version` or `python test_versioning.py`. This is validated during the `npm run dist`
```

BSd3v (approved):
💃

PR #328, opened 2024-09-30
bump to AG Grid version `32.3`

BSd3v:


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "dash-ag-grid",
-    "version": "31.3.1",
+    "version": "32.3.0",
```
ndrezn, line 3: Shouldn't have this diff

package.json
```diff
@@ -25,17 +25,16 @@
...
     "license": "MIT",
     "dependencies": {
         "@emotion/react": "^11.11.3",
         "@emotion/styled": "^11.11.0",
-        "ag-grid-community": "31.3.4",
-        "ag-grid-enterprise": "31.3.4",
-        "ag-grid-react": "31.3.4",
+        "ag-grid-community": "32.3.4",
+        "ag-grid-enterprise": "32.3.4",
+        "ag-grid-react": "32.3.4",
```
ndrezn, line 37: Is retaining the minor and patch versions deliberate? do we want to just update to the latest minor/patch versions?

package.json
BSd3v: We are bumping the major version, so that should be correct.

package.json
BSd3v: Yes, we didnt want things to randomly get out of sync.

package.json
ndrezn: Gotcha, missed that.

src/lib/fragments/AgGrid.react.js
```diff
@@ -1494,3 +1502,6 @@ DashAgGrid.propTypes = {parentState: PropTypes.any, ..._propTypes};
 
 export const propTypes = DashAgGrid.propTypes;
 export const defaultProps = DashAgGrid.defaultProps;
+
+var dagfuncs = (window.dash_ag_grid = window.dash_ag_grid || {});
+dagfuncs.useGridFilter = useGridFilter;
```
ndrezn, line 1507: What's `dagfuncs`?

BSd3v: A way to pass the internal api things to the app developer. For example, this passes `useGridFilter` to be used in the difference of `useImperativeHandler` between 31 and 32. This way they can use filters like they were the component developer.

T4rk1n (approved):
:woman_dancing: Looks good.

PR #331, opened 2024-10-07
creating `customSetProps` to handle `setProps`

BSd3v:
creating `customSetProps` to handle `setProps` when component no longer in dash tree

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source D
    - [#312](https://github.com/plotly/dash-ag-grid/issues/312) fixing issue where `scrollTo` was defaulting to not reset the value
      - to maintain scroll position during a grid rerender, be sure to use `getRowId`
    - fixing side issue where `cellDoubleClicked` was forcing the grid to rerender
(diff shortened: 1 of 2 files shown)


ndrezn:
Is there a specific issue that this change resolves?

BSd3v:
There are a couple of cases that lead to this:
- when a component is removed entirely, as in a page container removes the element or removed as a child from a dom (eg test_cs002_column_state)
  - this leads to `setProps` from internal component updates not merging with the grid component in dash and results in a dash renderer error for passing an object to an expected component
- when a component is unmounted and the underlying grid can no longer perform actions, this pertains to updates when no longer in the DOM tree (eg tabs that dont keep the component in the DOM)
  - this leads to an error inside the component where the state (mounted) and props are out of sync and it tries to apply changes to the grid even though it is currently unmounted. Often seen as JS errors.

gvwilson:
@T4rk1n can you please have a look at this? Once it is reviewed and merged, @BSd3v can do a new dash-ag-grid release.

T4rk1n (approved):
LGTM :woman_dancing:

gvwilson:
Thanks @T4rk1n - @BSd3v please go ahead and push the button.

PR #340, opened 2024-11-21
Run pytest after package installation

ndrezn:
For some reason on `release.yml`, tests were being run before packages were installed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,6 +29,7 @@ jobs:
           source .venv/bin/activate
           uv pip install ".[dev]"
           npm i
+          npm run test
           npm run dist
           npm pack && mv *.tgz dist/
         shell: bash
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,19 +19,36 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.10"
-      - name: Setup uv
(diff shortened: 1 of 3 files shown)


gvwilson (approved):

package.json
BSd3v, line 1: Eh, I dont know about this.

This assumes that the dev wants to run tests before creating a distro. For example, what if I wanted to install it onto an app using `npm run dist`, I'd have to sit through the tests again.

.github/workflows/release.yml
```diff
@@ -24,15 +24,10 @@ jobs:
...
-          uv pip install ".[dev]"
-          pytest --headless
       - name: Generate distribution
         run: |
           source .venv/bin/activate
-          uv pip install -r requires-dev.txt -r requires-install.txt
+          uv pip install ".[dev]"
           npm i
           npm run dist
           npm pack && mv *.tgz dist/
```
BSd3v, line 33:
```
npm i
npm run dist
pytest --headless
npm pack && mv *.tgz dist/
```

I'd just add the test there.

BSd3v (changes requested):

package.json
ndrezn: I'll make it an optional step. I don't think we ever want to allow generating a dist without passing tests but we can enforce it in CI anyway.

BSd3v (approved):
💃

PR #346, opened 2024-12-02
Fix Styling for Nodes

BSd3v:
adjusting styling to check if the `node` exists, instead of testing for `data`, this fixes an issue with `rowGrouping` not passing through the styling conditions

### DIFF
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -1125,7 +1125,7 @@ export default class DashAgGrid extends Component {
             return (params) => {
                 for (const {test, style} of tests) {
                     if (params) {
-                        if (params.data) {
+                        if (params.node) {
                             if (test(params)) {
                                 return style;
                             }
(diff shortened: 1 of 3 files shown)


src/lib/fragments/AgGrid.react.js
```diff
@@ -1125,7 +1125,7 @@ export default class DashAgGrid extends Component {
             return (params) => {
                 for (const {test, style} of tests) {
                     if (params) {
-                        if (params.data) {
+                        if (params.node.id && params.node.id !== null) {
```
gvwilson, line 1128: why does this fix it?

BSd3v: When using enterprise and row grouping, the `data` is not present on the rows that are grouped. This was causing an issue when trying to apply styling based upon aggregate info.

However, removing `data` and just letting the values pass would break the app if the dev didnt have something like this in the condition: `params.data ? params.data[column] : null`.  This was especially the case when dealing with `infinite` scroll. 

Instead, by checking to see if the `node.id` exists and is not null allows an actual row to be tested, instead of rows that arent quite in existance.

gvwilson (approved):

ndrezn (approved):
👍 💃 Top-level object comparison isn't going to be sufficient... nice fix.

PR #355, opened 2025-02-06
Bump ag-grid to version 33.0.4

Godisemo:
This PR bumps ag-grid version to 33.0.4

I would also like to add a `theme` parameter to the `AgGrid` component as well, since this is the preferred way to style the grid now. Can I do that in this PR as well or is it better to save for a separate PR?


BSd3v:
No, there a breaking changes between each major release. We should step up 31, then 32, and then support both 32 and 33. 

Upstream AG grid has stated that they will support 32 and 33 because there are so many changes between the 2.

PR #368, opened 2025-04-10
Fix NPM publish step

ndrezn:
Properly fetch file name and pass to publish command.


BSd3v (approved):
💃 as long as it works.

PR #380, opened 2025-07-10
Refactor: Migrate to function components

KoolADE85:
This PR is a pure refactor from class to function components.

There is no functionality change. All automated tests continue to pass, and all example apps in `docs/` have been manually confirmed as working.

### DIFF
--- a/src/lib/fragments/AgGrid.react.js
+++ b/src/lib/fragments/AgGrid.react.js
@@ -1416,6 +1416,7 @@ export function DashAgGrid(props) {
             // Hydrate virtualRowData and finalize setup
             onFilterChanged(true);
             updateColumnState();
+            setColumnState_push(false);
         }
     }, [
         gridApi,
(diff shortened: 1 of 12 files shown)


AnnMarieW:
@KoolADE85 
This is looking good!   Any plans on moving away from `defaultProps`?

tests/test_cell_data_type_override.py
```diff
@@ -77,6 +77,6 @@ def test_cd001_cell_data_types_override(dash_duo):
         # test overriden dateString cell data type
         action.double_click(grid.get_cell(0, 1)).perform()
         date_input_element = dash_duo.find_element(f'#{grid.id} .ag-date-field-input')
-        date_input_element.send_keys("01172024" + Keys.ENTER)
+        date_input_element.send_keys("01012024" + Keys.ENTER)
 
-        grid.wait_for_cell_text(0, 1, "17/01/2024")
+        grid.wait_for_cell_text(0, 1, "01/01/2024")
```
BSd3v, line 82: Why'd you change this?

This should have formatted correctly, based upon the data type override.

src/lib/fragments/AgGrid.react.js
```diff
@@ -219,1282 +306,1370 @@ export default class DashAgGrid extends Component {
...
+        props.deleteSelectedRows,
+        paginationGoTo,
+        scrollTo,
+        resetColumnState,
+        exportDataAsCsv,
+        selectAll,
+        deselectAll,
+        deleteSelectedRows,
+        customSetProps,
+    ]);
```
BSd3v, line 1411: this seems like overkill, and this may even be uneccessary considering the other `useEffect` are listening for the state change on `gridApi` as well.

tests/test_cell_data_type_override.py
BSd3v: January 1st will always present the same way.

BSd3v (approved):
Looks good to me. All tests are passing.

PR #381, opened 2025-07-15
Feature: AG Grid 33

KoolADE85:
This PR upgrades AG Grid to version 33.


src/lib/components/AgGrid.react.js
```diff
@@ -1,6 +1,17 @@
...
+    AllCommunityModule,
+    ModuleRegistry,
+    provideGlobalGridOptions,
+} from 'ag-grid-community';
+
+// Register all community features
+ModuleRegistry.registerModules([AllCommunityModule]);
+
+// Mark all grids as using legacy themes
+provideGlobalGridOptions({theme: 'legacy'});
```
BSd3v, line 14: Not sure if we want to force people to do this.

BSd3v:
Also, we should look at this, if this PR is good, should probably do theirs. :)

https://github.com/plotly/dash-ag-grid/pull/355

src/lib/components/AgGrid.react.js
KoolADE85: OK, that's a great callout. I removed this and the other legacy theme imports in 0d79034
It does change the default theme, but I guess that's ok too. It's AG Grid's defaults which probably makes more sense than us setting our own, different defaults.

KoolADE85:
> Should make sure there is a way to set the theme to legacy though. So that people can opt for not having to create the theme.

OK, this gets a little tricky since AG Grid doesn't support importing legacy theme CSS files (as we were previously) AND setting a modern theme object.

So in my update, I propose that anyone who wants to keep using a legacy CSS theme, to import the CSS by hand (we won't automatically import, for example, `ag-grid.css`. An example of this is in `tests/examples/themes_legacy.py`.

Our codebase going forward, would support AG Grid's modern theme system by default. See `tests/examples/themes.py`.

This approach seems right because, in case of any errors/warnings, the message that AG Grid outputs in the console matches what users would need to do to their dash app in order to get the grid looking right.

BSd3v:
> > Should make sure there is a way to set the theme to legacy though. So that people can opt for not having to create the theme.
> 
> OK, this gets a little tricky since AG Grid doesn't support importing legacy theme CSS files (as we were previously) AND setting a modern theme object.
> 
> So in my update, I propose that anyone who wants to keep using a legacy CSS theme, to import the CSS by hand (we won't automatically import, for example, `ag-grid.css`. An example of this is in `tests/examples/themes_legacy.py`.
> 
> Our codebase going forward, would support AG Grid's modern theme system by default. See `tests/examples/themes.py`.
> 
> This approach seems right because, in case of any errors/warnings, the message that AG Grid outputs in the console matches what users would need to do to their dash app in order to get the grid looking right.

We can still support these files and not import them directly into the grid the same way. There are a couple of repos that do this in order to make it easier to import these files that are otherwise dependencies to make the package load properly.

KoolADE85:
Alright, I'm giving this a try: lazy load the css for legacy themes if the user has specified a `className` (and no theme). Otherwise, default to AG Grid's modern theme system.
There's quite a bit of extra code to keep this backwards-compatible API, but it does achieve the goal of allowing users to upgrade without needing to refactor their theming.
What do you think?

.gitignore
BSd3v, line 1: these should be tracked and pushed

dash_ag_grid/__init__.py
```diff
@@ -9,21 +9,18 @@
 # noinspection PyUnresolvedReferences
 from ._imports_ import *
 from ._imports_ import __all__
+from dash_ag_grid import themes
```
BSd3v, line 12: Does this make more sense to be imported into the `_imports_` rather than adding another line here?

KoolADE85: I think _imports_ is auto-generated, so I'm not sure how to safely add this in there!

dash_ag_grid/themes.py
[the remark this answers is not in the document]
KoolADE85: It does work in the example app (`tests/examples/themes_legacy.py`)!
But I suppose it also puts us at the whims of jsdelivr 😅

.gitignore
BSd3v: Oh wait, sorry, I got it in reverse, carry on.

dash_ag_grid/themes.py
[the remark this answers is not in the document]
BSd3v: Maybe something with just a `getBoundClientRect` to the expected width of the grid based upon window size.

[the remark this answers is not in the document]
KoolADE85: OK, I added a test that ensures both the base stylesheet and the given theme's stylesheet is applied (and confirmed that this test fails if there are bad URLs in `themes.py`).

BSd3v:
> Also, we should look at this, if this PR is good, should probably do theirs. :)
> 
> #355

It seems like #355 is covered by this current PR, with a couple of other features.

KoolADE85:
> > Also, we should look at this, if this PR is good, should probably do theirs. :)
> > #355
> 
> It seems like #355 is covered by this current PR, with a couple of other features.

Yes, I think that one can be closed in favour of this one - we have theme support now here, plus latest patch release of `33`.

BSd3v (approved):
LGTM 💃

PR #384, opened 2025-07-16
Enforce a locale for cell formatting test

KoolADE85:
This PR fixes the cell formatting test which can fail if the browser's locale is not `en-US`.


BSd3v (approved):
💃 
LGTM

PR #386, opened 2025-07-18
Bugfix: race condition in paginationGoto prop

KoolADE85:
This PR fixes a flaky test (and real world problem) where pagination might not be set as per the prop, depending on the order of operations in the component.


src/lib/fragments/AgGrid.react.js
BSd3v, line 1: Looks like I missed that one when I was removing the dupes

PR #391, opened 2025-07-24
removed deprecated defaultProps

### LINKED ISSUE #390: Remove `defaultProps`
Remove the `defaultProps` to get rid of the console warning.

@AnnMarieW 🥇

AnnMarieW:
Closed #390 

This fixes the console error about defaultProps being removed in React 19


AnnMarieW:
> Could we do something like what you are doing with the persistence:

I don't think so - defaultProps is deprecated.

PR #400, opened 2025-08-22
minor tweak to column state chain for updates

BSd3v:
This should also be applied to v32.

Fixes issue here: https://community.plotly.com/t/custom-function-as-dash-row-style/93772/10?u=jinnyzor


KoolADE85 (approved):
Working well on my end. Great fix! 💃

PR #408, opened 2025-09-18
fixes issue with sticky columnState when updating columnDefs alone

BSd3v:


KoolADE85 (approved):
Works well on my end! 💃

Issue #410, opened 2025-10-23
Server side row model does not work


anujdhillxn:
Hi,

On investigation of the codebase, i found that there is no handling for `serverSide` row model. It is available only for the infinite row model.

serverSide row model is necessary to enable on-demand fetching of row groups.
We can enable this by handling this in a similar fashion as done in the 'infinite' row model.

PR #411, opened 2025-10-23
Added support for server side row model to enable server side row grouping, similar to infinite row model

anujdhillxn:


anujdhillxn:
Fixes https://github.com/plotly/dash-ag-grid/issues/410

anujdhillxn:
Hi @gvwilson. Thanks for the response. I have edited the PR title and the attached issue to be more descriptive.

BSd3v:
@anujdhillxn,

I am open to this, however, can you provide some more types of tests. 

For example, we need to have grouping (this is what you have), filters and sorting, just like how we have it in the infinite model.

We had initially avoided this due to complexity and the logic behind setting it up, also had issues with creating the payload to the dash server for the necessary details.

Adir207:
Hi @anujdhillxn,
Does this makes the usage of `serverSide` row model similar to the `infinite` model? Meaning we can utilize the `getRowRequest` in a callback similarly to what we do for the `infinite` model? Thanks!

PR #419, opened 2025-12-13
Add onRowDoubleClicked event

saddfox:
Added `onRowDoubleClicked` event, following example in https://github.com/plotly/dash-ag-grid/pull/201
[AG Grid docs ](https://www.ag-grid.com/javascript-data-grid/grid-events/#reference-selection-rowDoubleClicked)


saddfox:
Okay, I understand. Thanks

PR #424, opened 2025-12-19
updating changelog for release

BSd3v:


BSd3v:
@KoolADE85,

Idk if we want to name it this or not, @ndrezn and I were thinking of just a full release as the only hold up are the docs and this has been passing tests.

KoolADE85 (approved):
Looks good on my end. @LiamConnors do you have any objections to removing the `rc` tag here ahead of an official docs update?

PR #428, opened 2025-12-30
Allows for `createPart` to be used to craft parts to be utilized with `withPart`

BSd3v:
AG Grid is looking for a specific type when using the `withPart`, this allows a developer to create their own parts and configure properly for the grid to accept

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Links "DE#nnn" prior to version 2.0 point to the Dash Enterprise closed-source D
     - fixes [#416] (https://github.com/plotly/dash-ag-grid/issues/416)
     - fixes [#407](https://github.com/plotly/dash-ag-grid/issues/407)
 - [#412](https://github.com/plotly/dash-ag-grid/issues/412) fix "Multi-Column Filter not properly recognized in filterParams"
+- [#427](https://github.com/plotly/dash-ag-grid/issues/427) exposes `agGrid` from the community package for use in custom themes and functions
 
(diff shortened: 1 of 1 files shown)


AnnMarieW:
Should the js grid be imported automatically so you don't have to do it like this in all the apps?

```python

app = Dash(
    external_scripts=[
            f'https://cdn.jsdelivr.net/npm/ag-grid-community@{dag.grid_version}/dist/ag-grid-community.min.js'
    ]
)
```

src/lib/fragments/AgGrid.react.js
```diff
@@ -47,6 +47,7 @@ import {
     themeBalham,
     themeMaterial,
     themeQuartz,
+    createPart,
```
KoolADE85: Since we already have ag-grid-community imported, it seems like we could avoid the requirement to download it as an external script by supplying it as context for the function?

```
import * as agGrid from 'ag-grid-community';

... 

const context = {
  ...themes,
  agGrid,
  d3,
  ...etc
};
```

Then, in the app:

```
dagfuncs.customTheme = (theme, agGrid) => {
    return theme.withPart(agGrid.createPart(agGrid.colorSchemeDark))
}
```

So basically, instead of cherry-picking `createPart`, we could just send over the entire agGrid api?

BSd3v: It added 1kb, I could see it as worth it.

KoolADE85 (approved):
This works well on my end!

PR #452, opened 2026-04-21
fix #451

### LINKED ISSUE #451: [Bug] Broken `dash-ag-grid==35.2.0` for certain `dataTypeDefinitions` configuration
The following minimal reproducible example works fine with the `dash-ag-grid==33.3.3`, but it crashes with the latest `dash-ag-grid==35.2.0`. 

```py
import dash
from dash import html
import dash_ag_grid as dag
import plotly.express as px

df = px.data.iris()

app = dash.Dash(__name__)

app.layout = html.Div(
    [
        dag.AgGrid(
            columnDefs=[{"field": c} for c in df.columns],
            dashGridOptions={
                "dataTypeDefinitions": {
                    "number": {
                        "columnTypes": "rightAligned",
                    }
                }
            }
        )
    ]
)

if __name__ == "__main__":
    app.run(debug=True)
```

Here's what I get when I open the app with the `dash-ag-grid==35.2.0`:

https://github.com/user-attachments/assets/7820adfa-8c80-4f04-8b1e-a46494c6e41b

I guess the bug is introduced with this PR -> https://github.com/plotly/dash-ag-grid/pull/440/, but not fully sure. 

We in Vizro team would like to stay updated with the latest `dash-ag-grid`, but our default configuration for `dag.AgGrid` objects in Vizro contains the configuration from the example above. So for now, we're sticked with the `dash-ag-grid=33.3.3`.

Thanks in advance, and let us know if you need any further information or help 🚀

BSd3v:
fixes #451
---

wraps to test for object before trying to parse


AnnMarieW (approved):
LGTM 🙂

petar-qb:
Thank you @BSd3v on such a quick fix! 🚀

PR #469, opened 2026-06-24
fix tests

AnnMarieW:
Fix tests -  use dataset from plotly repo rather than ag grid


BSd3v:
@AnnMarieW,

When we were looking at this, I was thinking of something like a utils function that would be used across the board, this could also potentially allow for providing cached versions of the data as well. This could allow us to switch data sources easily.

There is still a `ag-grid` site reference for a couple of tests.

BSd3v (approved):

PR #470, opened 2026-06-27
Support processUnpinnedColumns

### LINKED ISSUE #468: Support processUnpinnedColumns as a function
It would be nice to have support for the global option `processUnpinnedColumns` which is currently missing in `GRID_MAYBE_FUNCTIONS`.

https://www.ag-grid.com/javascript-data-grid/grid-options/#reference-columnPinning-processUnpinnedColumns

datenzauberai:
This would close #468 

It basically just adds processUnpinnedColumns to src/lib/utils/propCategories.js and adds a test for this.

Please feel free to make changes.


tests/utils.py
```diff
@@ -44,11 +46,40 @@ def _wait_for_count(self, selector, expected, description):
             raise ValueError(f"found {len(els)} {description}, expected {expected}")
 
     def wait_for_pinned_cols(self, expected):
-        # TODO: is there a pinned right?
+        self.wait_for_pinned_column_count(expected, pin_state="left")
+
+    def _header_class_for_pin_state(self, pin_state: Literal["left", "right", "scrolling"]):
```
datenzauberai, line 51: The header classes will change with AgGrid 36 (.ag-grid-pinned-left-cells .ag-grid-scrolling-cells .ag-grid-pinned-right-cells) so I thought it would be good to already locate this in one method.

tests/utils.py
```diff
@@ -44,11 +46,40 @@ def _wait_for_count(self, selector, expected, description):
             raise ValueError(f"found {len(els)} {description}, expected {expected}")
 
     def wait_for_pinned_cols(self, expected):
-        # TODO: is there a pinned right?
+        self.wait_for_pinned_column_count(expected, pin_state="left")
```
datenzauberai, line 49: I frowned upon refactoring all calls to `wait_for_pinned_column_count`, so kept this in place and removed the TODO, because the other methods now would support left/right/scrolling.

tests/test_column_pinning.py
```diff
@@ -0,0 +1,49 @@
...
+                }
+            ),
+        ],
+        style={"margin": 20},
+    )
+    dash_duo.start_server(app)
+
+    grid = utils.Grid(dash_duo, "grid")
+
+    grid.wait_for_pinned_column(col_id="COL_0", pin_state="left")
```
datenzauberai, line 53: This would fail without unpinAllButFirstColumn, because default AgGrid behaviour would be to start unpinning at COL_0.

AnnMarieW:
Hi @datenzauberai 

Thanks so much for the PR and the updated tests :tada:

I'm going to switch this to the v35 branch, which should fix the unrelated tests that are failing.
What version are you using?  Do you need this applied to an earlier version?

datenzauberai:
> I'm going to switch this to the v35 branch, which should fix the unrelated tests that are failing.

Perfect.

> What version are you using? Do you need this applied to an earlier version?

We're on 33, but we're not blocked by this, because we workaround using a clientside callback to set the parameter. Hence, any release would just be super fine for us.

tests/utils.py
BSd3v: Sounds good to me.

tests/utils.py
BSd3v: You could always default to pin_state=`left` so you could replace.

Also, I'm not sure if we want to just count the number of pinned columns, vs checking against their actual values?

tests/test_column_pinning.py
BSd3v, line 1: What happens to this test if the screen size is smaller?

We've run into this issue a few times, so it might be work scaling the window to a measure so that this doesnt arbitrarily fail.

tests/utils.py
datenzauberai: If you are fine with extending this PR I could:

- remove wait_for_pinned_cols and wait_for_viewport_cols
- refactor test cases that used to call wait_for_pinned_cols or wait_for_viewport_cols to explicitly check for pin state for the columns where applicable or call wait_for_pinned_column_count

Would you like to have that change?

tests/test_column_pinning.py
datenzauberai: Smaller screen sizes (width < 2000px):

test will work, because the columns are 200px wide, there are 10 columns, all columns are pinned, AG Grid will trigger processUnpinnedColumns.

Bigger screen sizes (width > 2000px):

test will fail, because the pinned columns do not overflow.

If you prefer, I can parametrize the test with window sizes and determine the number of columns depending on the window size, e.g.

```python
@pytest.mark.parametrize("window_size", [(800, 600), (1280, 720), (3840, 2160)])
def test_cd001_process_unpinned_columns(dash_duo, window_size):
    width, height = window_size
    dash_duo.driver.set_window_size(width, height)
    column_width = 200
    column_count = int(width / column_width) + 1
```

BSd3v: You dont need to parametrize it, just set the window height so that it is going to work regardless of screen size / zoom.

tests/utils.py
BSd3v: I think we can wait for this, I do think that you can at least default for the pin_state to left.

BSd3v:
@datenzauberai 

We need a changelog entry.

tests/test_column_pinning.py
datenzauberai: Done

tests/utils.py
datenzauberai: Okay, I kept the existing test helpers for now and just default pin_state to left.

datenzauberai:
> @datenzauberai
> 
> We need a changelog entry.

Done.

BSd3v (approved):
💃

datenzauberai:
Sorry for the dumb question, but as I haven't contributed to this project before... anything else needed from my side?

BSd3v:
Nope, wasnt sure if you get the green button or not. 😄

PR #476, opened 2026-07-15
Switch npm publish to trusted publishing (OIDC)

ndrezn:
## Why

Rather than rotate a credential that will expire again, this switches to **trusted publishing** via GitHub OIDC — the same approach `pypi-publish.yml` already uses for PyPI. No long-lived secret to expire or rotate.

## Changes

- Drop `NODE_AUTH_TOKEN` / `secrets.NPM_TOKEN` from the publish step (auth now via OIDC)
- Drop the explicit `--provenance` flag (provenance is automatic with trusted publishing)
- Add an npm upgrade step — trusted publishing OIDC requires **npm >= 11.5.1**, newer than the version bundled with Node 20

`id-token: write` and `environment: name: npm` were already present, so no permission changes are needed.

npmjs.com has been updated.


BSd3v (approved):
💃

PR #479, opened 2026-07-16
Fix npm publish: bump Node to 24, pin npm to 11.x (v34)

ndrezn:
## Why

The `v34.3.0` release run failed at the "Upgrade npm for trusted publishing" step with `EBADENGINE`: `npm install -g npm@latest` now floats to **npm 12**, which requires Node >= 22.22/24.15, but the job runs Node 20.

## Fix

- Bump the publish job to **Node 24** (also clears the Node 20 deprecation warning).
- Pin the upgrade to **`npm@11`** (satisfies the `npm >= 11.5.1` trusted-publishing requirement without drifting into an incompatible major).

Backport of #478 (into `v35`) to the `v34` branch. Once merged, the `v34.3.0` publish can be re-run via `workflow_dispatch`.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -24,12 +24,13 @@ jobs:
     steps:
       - uses: actions/setup-node@v4
         with:
-          node-version: '20.x'
+          node-version: '24.x'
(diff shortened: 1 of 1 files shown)


BSd3v (approved):
💃

